PR #5, opened 2016-07-07
replace old style string formatting with new format

kmario23:
Replaced the old style of string formatting using percents with new "format", which is kinda neat.


sdispater:
The tests are broken.

It seems that you have unnecessary parenthesis, for instance:

```python
return '{}_{}'.format((m.group(1).lower(), m.group(2).lower()))
```
should be

```python
return '{}_{}'.format(m.group(1).lower(), m.group(2).lower())
```

kmario23:
@sdispater Sorry about that! I fixed it and the build is successful now.

sdispater:
No worries 😃 
Thanks for the pull request

PR #11, opened 2016-07-08
Added install command to README

willtheorangeguy:
Thought it would be helpful for end users if we added the install command at the end so that they don't necessarily have to visit the website.


sdispater:
You can just take the Installation section of the official documentation. Pendulum is also available on other systems, not only Windows, so it doesn't make sense to reference it here.

willtheorangeguy:
Okay. That's fine.

PR #25, opened 2016-08-18
Fixing lang/pl.py - changing before and after

MichalTHEDUDE:
after and before definitions should be switched :)


sdispater:
Thanks :-)

PR #31, opened 2016-08-23
Add function and decorator to recursively replace all stdlib datetime instances with Pendulum instances, including those returned by functions and by generator objects

iv597:
At SpotOn we're trying to migrate from a bunch of internally written datetime util functions to Pendulum, and wanted a quick way to convert large structures of naive datetime objects from our database without converting them one by one. This is, in my opinion, a cleaner solution than monkey patching PyMongo or other similar shenanigans.

As the tests show, this can be used as a standalone function or as a function decorator to simply one-liner refactor existing codebases. It will wrap just about any source of datetime objects I could think of (if I've missed any, please let me know and I'll update the branch - I write about a 6:1 ratio of JS to Python here and don't always remember all the types!), including those generated by generator functions or function/lambda calls.

Wasn't sure where to put this kind of thing - it feels a little too invasive to me to want to merge it straight into the main library, but too small in scope to release to PyPy individually without at least running it by the upstream project first. Thus, I created a contrib folder and threw it in there. However, if you feel this functionality is not suited to be shipped with Pendulum upstream at all, we're still more than willing to release it as a standalone module on PyPy.


sdispater:
Thanks for taking the time to make this pull request. It looks really clean :-)

I can see how this may useful when you want to migrate easily an existing codebase. However, I think it would be better as an external package that you would bring when necessary.

And I don't think there is something as "too small to release on PyPi" ;-) I mean, if it is useful there is no reason to not share it with the community.

iv597:
Noted! We'll get this packaged up. Thanks for taking the time to review!

iv597:
Done! Closing this PR.

https://github.com/SpotOnInc/pendulumify
https://pypi.python.org/pypi/pendulumify

PR #35, opened 2016-08-24
Refactor `Parser` into `Loader`

### LINKED ISSUE #34: Loader.load portability broken for some pytz installs
I don't know the full details that make my platform not work with `Loader.load` but I'm using a Fedora Core 24 box with its stock python3.5 interpretor.  I believe that the `pytz` install I have was installed by pip3 but I'm not 100% sure.

When I run tests most of them fail because `Loader.load` is trying to open zoneinfo files in a location that they don't exist.  Looking at the code it looks like this is a possibly a non portable way of getting zoneinfo since it makes assumptions about where the zoneinfo files should live.

I tweaked the exception to print the file location to illustrate..

```
======================================================================
ERROR: wrap_with_test_now (tests.tz_tests.test_timezone.TimezoneTest)
----------------------------------------------------------------------
  File "/home/mayfield/project/pendulum/tests/__init__.py", line 13, in setUp
    LocalTimezone.set_local_timezone(timezone('America/Toronto'))
  File "/home/mayfield/project/pendulum/pendulum/tz/__init__.py", line 16, in timezone
    return Timezone.load(name)
  File "/home/mayfield/project/pendulum/pendulum/tz/timezone.py", line 50, in load
    default_transition_type) = Loader.load(name)
  File "/home/mayfield/project/pendulum/pendulum/tz/loader.py", line 28, in load
    raise TypeError(filepath)
TypeError: /usr/lib/python3.5/site-packages/pytz/zoneinfo/America/Toronto
```

In that case there is not a `zoneinfo` sub dir in `/usr/lib/python3.5/site-packages/pytz`.  But on the same machine this works fine..
```
>>> pytz.timezone('America/Toronto')
<DstTzInfo 'America/Toronto' LMT-1 day, 18:42:00 STD>
```

mayfield:
* `Parser.parse` -> `Loader.load`
* `Parser._parse -> `Loader._load`
* Use `pytz.open_resource` for zone file loading to fix pytz installs
  using shared zoneinfo files.
* Fix for python2 setup and test (backport `FileNotFoundError` name).

Tested on Fedora Core 24 (linux), python2.7 (distro), python3.5
(distro)

Fixes #34


sdispater:
Thanks. Looks clean :-)

Just a quick question though: why did you put

```python
try:
    FileNotFoundError
except NameError:
    FileNotFoundError = IOError  # py2k
```

in `setup.py`?

I just tested without and it works fine, but I'm on OSX so there might be some issues on Linux systems that require this.

mayfield:
FileNotFoundError is only available on python3.  Actually only on 3.3+ I think.  How it would be available otherwise is beyond me.  Here is output from my osx box...

```
:; python
Python 2.7.12 (default, Jul 30 2016, 08:56:53) 
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> FileNotFoundError
NameError: name 'FileNotFoundError' is not defined
```

Same error.  I ran into this when doing `python ./setup.py test` from linux btw.

sdispater:
I didn't know and I was not getting the error because I'm not using `python setup.py test` but I directly invoke `py.test`

PR #39, opened 2016-08-30
Resolve an AttributeError when calling between

iv597:
This seemed to be caused by passing two Pendulum objects created by `pendulum.instance(... datetime ...)` to `pendulum.now().between(...)`. Removing this explicit conversion prevents a double conversion caused by the `__gt__` method, which raised an AttributeError (because, of course, native DT objects, already extracted, will not contain a `_datetime` attribute).

This bug was at least found in pendulum 0.5.4 on python 2.7, my local dev environment for my current project.

### DIFF
--- a/pendulum/pendulum.py
+++ b/pendulum/pendulum.py
@@ -1166,9 +1166,6 @@ class Pendulum(datetime.datetime, TranslatableMixin):
 
         :rtype: bool
         """
-        dt1 = self._get_datetime(dt1)
-        dt2 = self._get_datetime(dt2)
-
         if dt1 > dt2:
             dt1, dt2 = dt2, dt1
 
--- a/tests/pendulum_tests/test_comparison.py
+++ b/tests/pendulum_tests/test_comparison.py
@@ -2,6 +2,7 @@
 
 import pytz
 from datetime import datetime
+from time import sleep
 from pendulum import Pendulum
 
 from .. import AbstractTestCase
@@ -293,6 +294,15 @@ class ComparisonTest(AbstractTestCase):
 
         self.assertFalse(d1.between(d3, d2, False))
         self.assertFalse(d1.between(d5, d4, False))
+
+    def test_between_issue_39(self):
+        old = Pendulum.instance(datetime.utcnow())
+        sleep(0.2)
+        mid = Pendulum.now('UTC')
+        sleep(0.2)
+        new = Pendulum.instance(datetime.utcnow())
+
+        self.assertTrue(mid.between(old, new))
 
     def test_min_is_fluid(self):
         d = Pendulum.now()


sdispater:
Thanks.

Could you just add a test that covers this error?

iv597:
I'll get that patched up shortly.

iv597:
@sdispater Test added and passes (at least in Python 2, I'll wait for Travis to get back to me about Py3)

sdispater:
Thanks.

However, the test you provided does not triggers any error in my case on the latest `0.5.4` version, it properly returns `True`.

Are you sure that the error is caused by the `_get_datetime()` method?

iv597:
@sdispater Interestingly the test does seem to pass with 0.5.4, but it's the closest test I can get to the code we're running without giving you a dump of our MongoDB structure, too.

Here's a sampling of the traceback I get on 0.5.4. This PR makes this error disappear.

```python
/usr/local/lib/python2.7/dist-packages/pendulum/pendulum.pyc in between(self, dt1, dt2, equal)
   1174 
   1175         if equal:
-> 1176             return self >= dt1 and self <= dt2
   1177 
   1178         return self > dt1 and self < dt2

/usr/local/lib/python2.7/dist-packages/pendulum/pendulum.pyc in __ge__(self, other)
   1148 
   1149     def __ge__(self, other):
-> 1150         return self._datetime >= self._get_datetime(other)
   1151 
   1152     def __lt__(self, other):

/usr/local/lib/python2.7/dist-packages/pendulum/pendulum.pyc in _get_datetime(self, value, pendulum)
   2152 
   2153         if isinstance(value, Pendulum):
-> 2154             return value._datetime if not pendulum else value
   2155 
   2156         if isinstance(value, datetime.datetime):

AttributeError: 'datetime' object has no attribute '_datetime'

In [3]: import ipdb; ipdb.pm()
> /usr/local/lib/python2.7/dist-packages/pendulum/pendulum.py(2154)_get_datetime()
   2152 
   2153         if isinstance(value, Pendulum):
-> 2154             return value._datetime if not pendulum else value
   2155 
   2156         if isinstance(value, datetime.datetime):

ipdb> value
datetime(2016, 8, 31, 17, 0, tzinfo=<TimezoneInfo [UTC, 0, False]>)
```

sdispater:
That's quite strange. If `value` is a datetime it should not go into this part of the code.

iv597:
@sdispater To clarify, I'm calling `between` on a Pendulum object, passing in two Pendulum objects created through `pendulum.instance()` - done by running a PyMongo result through [pendulumify](https://github.com/SpotOnInc/pendulumify). Pendulum's `between` function extracts the raw datetime from the two pendulum objects, but does not skip this block of code, thus raising the `AttributeError`.

If there's anything I can do or provide to help debug this, I'd be more than willing to do what I can. If you'd rather we opened an issue ticket to track the bug separately from my fix, that's cool too.

sdispater:
I can't figure out where the problem comes from but your solution seems good to me since it avoids the double call to `_get_datetime()`.

PR #47, opened 2016-09-07
Add support for instancing tz-aware datetime objects

iv597:
This is confirmed (by included unit tests) working with pytz'd datetimes, and also worked locally with some `psycopg2.tz.FixedOffsetTimezone` instances PostgreSQL gave us (which are obviously far out of scope of pendulum unit tests, and thus aren't included).

This makes some general assumptions about the type of stuff coming into the function:

* Pendulum TZ objects will never include `_utcoffset` or `_offset` properties. Seems legit - the Pendulum TZ classes currently don't expose anything of the sort.

* The incoming object *does* include one of those two properties, which contains a `datetime.timedelta` object (or something like it). We'll prefer to use `datetime.timedelta#total_seconds()` to create a Pendulum `FixedTimezone` if it's available (2.7+), but will fall back to calculating it ourselves if it's not. Shouldn't be an issue given Pendulum doesn't advertise 2.6 support, so feel free to remove that safety code if you'd like. It was borrowed from [here](https://stackoverflow.com/questions/3318348/how-can-i-extend-pythons-datetime-datetime-with-my-own-methods/14214646#14214646) anyway.

### DIFF
--- a/pendulum/pendulum.py
+++ b/pendulum/pendulum.py
@@ -199,6 +199,23 @@ class Pendulum(datetime.datetime, TranslatableMixin):
         """
         tz = dt.tzinfo or tz
 
+        if dt.tzinfo:
+            offset = None
+
+            if hasattr(dt.tzinfo, '_utcoffset'):
+                offset = dt.tzinfo._utcoffset
+            elif hasattr(dt.tzinfo, '_offset'):
+                offset = dt.tzinfo._offset
+
+            if offset is not None:
+                try:
+                    offset_seconds = offset.total_seconds()
+                except AttributeError:
+                    offset_seconds = float((offset.microseconds +
+                        (offset.seconds + offset.days * 24 * 3600) * 10**6)) / 10**6
+
+                tz = FixedTimezone(offset_seconds)
+
         return cls(
             dt.year, dt.month, dt.day,
             dt.hour, dt.minute, dt.second, dt.microsecond,
--- a/tests/pendulum_tests/test_construct.py
+++ b/tests/pendulum_tests/test_construct.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import os
+import pytz
 from datetime import datetime
 from pendulum import Pendulum
 from pendulum.tz import timezone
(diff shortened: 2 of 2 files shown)


sdispater:
Thanks :-)

Regarding `pytz` timezones, I already made a change on the `develop` branch (0b1017f0f52aaeddf610bb753995c1e20f9f2d66) to fully support them (They are converted to a full `pendulum` named timezone and not only a `FixedTimezone`).

Also, your PR could be improved by doing:

```python
offset = dt.utcoffset()
```
rather than:

```python
if hasattr(dt.tzinfo, '_utcoffset'):
    offset = dt.tzinfo._utcoffset
elif hasattr(dt.tzinfo, '_offset'):
    offset = dt.tzinfo._offset
```

iv597:
So, sounds like pytz is covered, but will it cover cases where the TZ is not provided by pytz? (notably this case of psycopg2, which gives something vaguely similar to a pytz object, but not quite)

Looks like we took pretty similar ways of solving this problem, so the code could be merged to handle both cases (your `develop` branch appears to simply throw an error if it's not Pendulum or PyTZ types - maybe keep the utcoffset/offset detection as another round of fallbacks before raising the exception?)

sdispater:
Yeah sure, that's what I meant, even though It  might not have been clear :-)

We can take your code for tzinfo other than those coming from pytz.

Regardless, a `datetime` instance always has a `utcoffset()` method returning a `timedelta`. So it's better to use this rather than accessing protected attributes.

iv597:
@sdispater In that case, let me know what you'd like me to do going forward, since my bizarre corner case isn't easily testable. I can leave the branch laying around and you're free to pick code out of it, or I can rebase off of develop and open a new PR, probably at the cost of unit tests (since I can't think of an easily testable case for these non-pytz timestamps).

sdispater:
To test other types of tzinfo, you could add a tzinfo provided by dateutil which is part of the dependencies.

sdispater:
I just pushed a commit on the `develop` branch that should handle any tzinfo: 552c34de70fe22606b7fdf34b8f45dc2c53a29cf

Issue #17, opened 2016-08-17
Feature proposal: Basic default intervals


JelteF:
I propose to add some very simple default intervals to the library. 
I propose to add an interval called `day`, `hour`, `minute` and so on to the lib. That way you can do stuff like this:

```python
import pendulum as pm

now_in_paris = pm.now('Europe/Paris')
tomorrow_in_paris = now_in_paris + pm.day
next_week_in_paris = now_in_paris + 7*pm.day
```

It's inspired by the Ruby way of adding dates according to this comment, but modified to be Pythonic: https://news.ycombinator.com/item?id=12301965

YBadiss:
what would be the advantage over doing

```python
tomorrow_in_paris = now_in_paris.add(days=1)
next_week_in_paris = now_in_paris.add(weeks=1)
```
as described in the readme?

If you want to use the `__add__` operator directly or pass in the value to add, couldn't you do something like
```python
from pendulum import Interval

now_in_paris = pm.now('Europe/Paris')
tomorrow_in_paris = now_in_paris + Interval(days=1)
next_week_in_paris = now_in_paris + Interval(weeks=1)
```

sdispater:
Even though I can see the appeal of Ruby's `Time.now + 1.day`, I don't think it brings that much over the existing API, except saving a few characters, and it loses the ability to create your durations programatically with a dict which I find useful.

JelteF:
@YBadiss Yes, it's not that these two methods do not work. It's simply that this method would be shorter and IMHO still easy to read and understand. Especially when subtracting something it would be shorter as `subtract` is quite long. 
Your second solution is also what I meant as the way to implement it, simply doing something like this (with the s suffixed names also added):
```python
week = weeks = Interval(weeks=1)
day = days = Interval(days=1)
seconds = seconds = Interval(seconds=1)
...
```

@sdispater I'm not proposing that we should replace the existing API, just add this very simple feature so it can be quickly used in cases where it would be useful. Isn't saving characters partly why this library is created in the first place?

PR #55, opened 2016-10-10
Add common intervals to the Interval class

flokli:
This is similar to the proposal in #17. Instead of having common intervals inside the ``pendulum`` module root, they are added as class variables to the ``Interval`` class.


sdispater:
Thanks for taking the time to put this PR together.

However, like said in the related issue, I don't think it brings anything worth integrating, especially if you make them class constants.

flokli:
@sdispater Could you then also close #17?

PR #64, opened 2016-10-29
Add validation for days of the week.

kleschenko:


pendulum/pendulum.py
```diff
@@ -787,6 +789,8 @@ def set_week_ends_at(cls, value):
 
         :type value: int
         """
+        if value not in range(7):
```
sdispater: I would do:

```python
if value not in cls._days:
```

pendulum/pendulum.py
```diff
@@ -787,6 +789,8 @@ def set_week_ends_at(cls, value):
...
+            raise ValueError('Invalid day of the week value: {}'.format(value))
```
sdispater: The message could be just:

```python
'Invalid day of the week: {}'.format(value)
```

pendulum/pendulum.py
kleschenko: Updated.

sdispater:
Thanks!

Issue #74, opened 2016-11-12
Off-by-one microsecond when parsing with certain timezones


dokai:
I found a curious behavior when parsing timestamps where the microsecond value is off by one depending on the timezone used.

As a minimal example I'm attempting to parse the string `2016-11-12T02:09:39.594000`.

Parsing with the default timezone works as excepted:
```python
In [3]: Pendulum.parse('2016-11-12T02:09:39.594000').microsecond
Out[3]: 594000
```
Using a particular timezone shows the off-by-one difference.
```python
In [4]: Pendulum.parse('2016-11-12T02:09:39.594000', 'America/Panama').microsecond
Out[4]: 593999
```
Several timezones (but not all by any means) have this behavior
```python
In [5]: len(pytz.common_timezones)
Out[5]: 436

In [6]: len({name for name in pytz.common_timezones if Pendulum.parse('2016-11-12T02:09:39.594000', name).microsecond != 594000})
Out[6]: 182
```
I don't think anything in the particular choice of time or timezone should result in a difference of a single microsecond?

sdispater:
It's not limited to `parse()`:

```python
>>> import pendulum
>>> pendulum.create(2016, 11, 12, 2, 9, 39, 594000, 'America/Panama').microsecond
593999
```

After checking, it seems that the wrong value is due to the C extension to calculate offsets. If you install `pendulum` without C extensions it works properly:

```
PENDULUM_EXTENSIONS=0 pip install pendulum
```

dokai:
It seems the difference between the Python and C versions is that Python rounds the value before casting to int which the C version does not.

C
```C
microsecond = (int64_t) (unix_time * 1000000) % 1000000;
    if (microsecond < 0) {
        microsecond += 1000000;
    }
```
Python
```python
microsecond = int(round(unix_time % 1, 6) * 1e6)
```
Using the `594000` microseconds from above we can see that due to the way floats are handled the modulo value varies:
```python
>>> 1.594000 % 1
0.5940000000000001

>>> 123.594000 % 1
0.5939999999999941
```
and without the `round()` call the Python version would be similar to C.
```python
>>> int((123.594000 % 1 ) * 1e6)
593999
>>> int(round(123.594000 % 1, 6) * 1e6)
594000
```

danilobellini:
[...]

PR #75, opened 2016-11-12
Fix issue with microsecond rounding in C extensions.

dokai:
As described in https://github.com/sdispater/pendulum/issues/74 there is
a difference in the behavior of the C extension and Python
implementation in terms of handling microseconds.

The problem is that a microsecond value passed in may have an off-by-one
difference for certain values due to how floating point numbers are
handled, unless rounding is performed.

```python
  >>> import pendulum
  >>> pendulum.create(2016, 11, 12, 2, 9, 39, 594000, 'America/Panama').microsecond
  593999
```

This commit adds the same rounding behavior to the C extension as used
in the Python version.

### DIFF
--- a/pendulum/_extensions/_helpers.c
+++ b/pendulum/_extensions/_helpers.c
@@ -3,6 +3,7 @@
 #include <Python.h>
 #include <stdint.h>
 #include <datetime.h>
+#include <math.h>
 
 #ifndef PyVarObject_HEAD_INIT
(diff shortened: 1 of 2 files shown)


sdispater:
Thanks for this PR.

However, I will wait to merge it because the implementation will most likely change to handle the microseconds better.

sdispater:
Commit 184b94a90ef9dcb1d6fcf4f4a890719fd43ff24e fixes the issue in a way that it doesn't break for high year value either.

dokai:
Thanks!

PR #81, opened 2016-11-25
Fix memory leak in local_time in C extension.

### LINKED ISSUE #80: Memory leak
If I run the following:

```python
import pendulum

while True:
    pendulum.now()
```

It will consume more and more memory until the interpreter dies. (The same does not happen with ```datetime.datetime.now()```.)

ntoll:
Does exactly what it says on the tin. Fixes #80.


ntoll:
This led to the solution: https://github.com/benhoyt/scandir/issues/6

PR #89, opened 2017-01-26
Fixing missing parentheses in method day_of_year

asrenzo:
Hello,

Sounds like you forgot some parentheses in method day_of_year.

First when using is_leap_year method and after when computing nb days.

Regards,

Laurent


pendulum/date.py
```diff
@@ -137,11 +137,11 @@ def day_of_year(self):
 
         :rtype: int
         """
-        k = 1 if self.is_leap_year else 2
+        k = 1 if self.is_leap_year() else 2
 
         return (
             (275 * self.month) // 9
-            - k * (self.month + 9) // 12
+            - k * ((self.month + 9) // 12)
```
sdispater: These parentheses are not needed since it's the same result either way. Unless I am missing something.

pganssle: I don't think it's the same result either way:

```python
>>> k = 12
>>> mon = 1
>>> k * ((mon + 9) // 12)
0
>>> k * (mon + 9) // 12
10
```

It may be that it's the same for all values k can currently hold I suppose, but if the version with parens is what you're expecting, it's probably cleaner to include the parens so that the semantics are better understood.

sdispater: Actually you are right: the parentheses are necessary since, otherwise, the result will be off by 1 day for non leap years.

Could you add a test covering this?

asrenzo:
Tests added

sdispater:
Thanks. However, the tests you added pass even without the fix so it won't prevent any regression.

I suggest using December 31 as the day to be sure to get a date that triggers the error since it currently returns 364 for a non-leap year instead of 365.

Also, could you put the tests in [tests/date_tests/test_getters.py](https://github.com/sdispater/pendulum/blob/master/tests/date_tests/test_getters.py) since it's where `day_of_year` and other getters tests are already?

sdispater:
Looks good to me :-) Thanks!

asrenzo:
You’re welcome.

Pendulum saved my life.

Laurent

Issue #101, opened 2017-03-16
parser doesn't seem to properly parse exif formatted datetime


jMyles:
Hey there.  Thanks for pendulum.

It seems that pendulum.parse() does the wrong thing for strings gleaned by PIL's ExifTags interface.

For example, on a Canon S-series camera, the string for the key 'DateTime" is: 
```'2016:12:26 15:45:28'```

By pendulum parses this into this Pendulum object:
```2017-03-16T11:21:27+00:00```

**Comment by sdispater** [comment]
Well, the thing is the string is not really standard (the `:` separator is mostly reserved for time information) so `pendulum` will fallback on the `dateutil` parser which produces the error:

```python
>>> from dateutil.parser import parse

>>> parse('2016:12:26 15:45:28')
datetime.datetime(2017, 3, 16, 15, 45, 28)
```

To make it work you have to use `from_format()`:

```python
>>> import pendulum

>>> pendulum.from_format('2016:12:26 15:45:28', '%Y:%m:%d %H:%M:%S')
# <Pendulum [2016-12-26T15:45:28+00:00]>
```

jMyles:
Roger.  I agree it's an odd format, but to the degree that `parse()` is generally applicable for datetime representations to be found in the wild: do you think that Exif data from Canon cameras is common enough to warrant special logic?

If so, I can work on a pull request.

SethMichaelLarson:
I'm +1 on this having dealt with EXIF data in the past.

emattiza:
+1 on the issue as well. implemented a unit test for the single case presented above, and a slight modification to the COMMON regex used in the parser for YYYY:MM:DD separators. from_format does look to be the appropriate parse option though fwiw. PR this?

sdispater:
Version `1.2.0` has just been released with this feature included.

PR #102, opened 2017-03-20
New Parser Behavior for ":" separator in date formats

emattiza:
Per sdispater/pendulum#101, implements exif format (: as separator in date formats) to regex parser

- [x] Add functional test for new expected parse functionality
- [x] Test COMMON regex alteration against prior test suite


sdispater:
I like it!

However, I don't see it as a bugfix but rather as a new feature (parsing of exif format), so it might be better to base this on the `develop` branch which is where new, non-breaking feature should be.

SethMichaelLarson:
Agreed.

emattiza:
cleaned up and rebased onto develop. Appreciate the feedback guys!

sdispater:
Thanks!

PR #103, opened 2017-03-24
Fix lang

eumiro:
Fix the Slovak translation (the former version comes from the Czech language)
Fix a typo in the Basque translation


sdispater:
Thanks!

PR #108, opened 2017-04-29
When creating without tzinfo, preserve fold value.

neonquill:
When creating a new Pendulum instance without passing in a TimezoneInfo object and using python 3.6, the internal fold parameter was not correct.  In this scenario, self._fold was being set based on the fold parameter in the underlying datetime object.  When using python3.6, this value exists, but is always set to 0 since Pendulum is not passing the fold parameter to the datetime constructor.

 - Changed self._fold to always be set based on the internal fold variable.
 - Added checks to the existing unit tests to verify that the fold parameter is being preserved.

All tests pass with python 3.6.1, did not test with any other versions.


sdispater:
Thanks a lot!

PR #134, opened 2017-07-10
Fix #133: Allow deepcopy on Timezone classes.

### LINKED ISSUE #133: TimezoneInfo cannot be deepcopied
When converting a Pendulum class to a `datetime` object, it still keeps the tzinfo of the type `TimeZoneInfo` of the Pendulum object. Then when trying to deepcopy that, it raises the TypeError seen below.
```python
from copy import deepcopy

import pendulum

pd = pendulum.parse('2017-07-11 17:19:18.686028+02:00')
dt = pd._datetime

deepcopy(pd)
deepcopy(dt)  # TypeError raised here.
```

Traceback

```
  File "/home/andreas/.pyenv/versions/project/lib/python3.5/copy.py", line 182, in deepcopy
    y = _reconstruct(x, rv, 1, memo)
  File "/home/andreas/.pyenv/versions/project/lib/python3.5/copy.py", line 291, in _reconstruct
    args = deepcopy(args, memo)
  File "/home/andreas/.pyenv/versions/project/lib/python3.5/copy.py", line 155, in deepcopy
    y = copier(x, memo)
  File "/home/andreas/.pyenv/versions/project/lib/python3.5/copy.py", line 223, in _deepcopy_tuple
    y = [deepcopy(a, memo) for a in x]
  File "/home/andreas/.pyenv/versions/project/lib/python3.5/copy.py", line 223, in <listcomp>
    y = [deepcopy(a, memo) for a in x]
  File "/home/andreas/.pyenv/versions/project/lib/python3.5/copy.py", line 182, in deepcopy
    y = _reconstruct(x, rv, 1, memo)
  File "/home/andreas/.pyenv/versions/project/lib/python3.5/copy.py", line 292, in _reconstruct
    y = callable(*args)
TypeError: __init__() missing 5 required positional arguments: 'tz', 'utc_offset', 'is_dst', 'dst', and 'abbrev'
```

I found this bug because of another issue I made https://github.com/romgar/django-dirtyfields/issues/101.

AndreasBackx:
I've added default values to the Timezone classes so that they can be deepcopied. I think it's appropriate to provide these as `datetime.tzinfo` also has an empty constructor.

### DIFF
--- a/pendulum/tz/timezone.py
+++ b/pendulum/tz/timezone.py
@@ -24,10 +24,10 @@ class Timezone(tzinfo):
     POST_TRANSITION = 'post'
     TRANSITION_ERROR = 'error'
 
-    def __init__(self, name, transitions,
-                 tzinfos,
-                 default_tzinfo_index,
-                 utc_transition_times):
+    def __init__(self, name=None, transitions=(),
+                 tzinfos=(),
+                 default_tzinfo_index=0,
+                 utc_transition_times=[]):
         """
         Constructor.
 
@@ -457,7 +457,7 @@ class FixedTimezone(Timezone):
 
     _cache = {}
 
-    def __init__(self, offset, name=None, transition_type=None):
+    def __init__(self, offset=0, name=None, transition_type=None):
         """
         :param offset: offset to UTC in seconds.
         :type offset: int
--- a/pendulum/tz/timezone_info.py
+++ b/pendulum/tz/timezone_info.py
@@ -5,7 +5,7 @@ from datetime import tzinfo, timedelta
 
 class TimezoneInfo(tzinfo):
 
-    def __init__(self, tz, utc_offset, is_dst, dst, abbrev):
+    def __init__(self, tz=None, utc_offset=0, is_dst=False, dst=None, abbrev='GMT'):
         """
         :param tz: The parent timezone.
         :type tz: Timezone
--- a/tests/pendulum_tests/test_behavior.py
+++ b/tests/pendulum_tests/test_behavior.py
@@ -2,6 +2,7 @@
 
 import pickle
 import pendulum
+from copy import deepcopy
 from datetime import datetime, date, time, timedelta
 from pendulum import Pendulum, timezone
 from .. import AbstractTestCase
@@ -103,3 +104,14 @@ class BehaviorTest(AbstractTestCase):
         dt = pendulum.create(1941, 7, 1, tz='Europe/Amsterdam')
 
         self.assertEqual(timedelta(0, 6000), dt.dst())
+
+    def test_deepcopy(self):
+        dt = pendulum.create(1941, 7, 1, tz='Europe/Amsterdam')
+
+        self.assertEqual(dt, deepcopy(dt))
+
+
+    def test_deepcopy_datetime(self):
+        dt = pendulum.create(1941, 7, 1, tz='Europe/Amsterdam')
+
+        self.assertEqual(dt._datetime, deepcopy(dt._datetime))


pendulum/tz/timezone.py
```diff
@@ -24,10 +24,10 @@ class Timezone(tzinfo):
...
     TRANSITION_ERROR = 'error'
 
-    def __init__(self, name, transitions,
-                 tzinfos,
-                 default_tzinfo_index,
-                 utc_transition_times):
+    def __init__(self, name=None, transitions=(),
+                 tzinfos=(),
+                 default_tzinfo_index=0,
+                 utc_transition_times=[]):
```
SethMichaelLarson: It's dangerous to use mutable default parameters. I'd suggest using `None` values and converting to the proper defaults within `__init__`.

SethMichaelLarson: Might want to add flake8 to be a part of Travis CI to catch things like this @sdispater

AndreasBackx: Good to know, I can change that in a follow-up commit. Let me know what else needs to be changed before it can be merged so I can make a new commit and squash.

SethMichaelLarson: Beyond this I think it'd be a good idea to implement `__copy__` and `__deepcopy__` which are [referenced in the Python docs for the `copy` module](https://docs.python.org/3/library/copy.html#copy.deepcopy) rather than adding default parameters. TBH I've never used these methods myself so some experimentation might be required.

pganssle:
Why would you want to be able to deepcopy a TimeZone object? Not saying I haven't done the same sort of thing before, but my understanding is that `tzinfo` subclasses are intended to be singletons or singleton-like.

SethMichaelLarson:
@pganssle If you're deep-copying something that holds onto an object that holds onto an object ... etc ... that eventually holds onto `tzinfo` you want `copy.deepcopy` to be able to give you a non-error back.

pganssle:
@SethMichaelLarson Hm.. But is this desirable if `tzinfo` objects are not in fact implemented as singletons but you're hoping for them to be?

One reason this can be a problem is that the logic for inter-zone comparison and intra-zone comparison is slightly different, and two things are considered different if `tz1 is not tz2`, even if `tz1 == tz2`.

SethMichaelLarson:
It's not desirable for them to be copied if they're meant to be singletons. People who know this library's internals should answer your question better than me. I'm just saying that this object should probably not raise an exception when it's copied otherwise it makes all objects that want to be copied and also hold onto this object into more difficult beasts.

sdispater:
@pganssle I agree that, ideally, tzinfo subclasses should be singletons. However, `pendulum` took an approach similar to `pytz` regarding timezones handling in its first stable version. This is something that will likely change in the next major version.

AndreasBackx:
@sdispater so what should be the eventual solution? Default `None` values, the implementation of `__copy__` and `__deepcopy__`, or both?

sdispater:
@AndreasBackx The `TimezoneInfo` class is not meant to be used directly. So I think the best approach here would be to implement `__copy__` and `__deepcopy__`.

SethMichaelLarson:
@AndreasBackx You don't need to implement `__copy__` because you're never directly copying a `TimezoneInfo` object. You might as well implement it for completeness sake.

AndreasBackx:
I found something while looking at the Django's source for their [`tzinfo` subclasses](https://docs.djangoproject.com/fr/1.8/_modules/django/utils/tzinfo/):

> Special requirement for pickling: A tzinfo subclass must have an \_\_init\_\_() method that can be called with no arguments, else it can be pickled but possibly not unpickled again. This is a technical requirement that may be relaxed in the future.

Snippet is from the [Python docs](https://docs.python.org/3.6/library/datetime.html#tzinfo-objects).

So we should follow this requirement and provide either default `None` or primitive type values or add `__getinitargs__(self)` that provides these. However I am not sure whether that method is used by deepcopy.

sdispater:
@AndreasBackx This is for pickling which has different requirements.

So we can go with only implementing `__deepcopy__` I think.

AndreasBackx:
@sdispater alright. I'm currently on holiday and get back at the beginning of August so I won't be able to make a new commit until then. If I haven't committed anything by then (let's say the first week of August), don't be afraid to ping me.

bolkedebruin:
@AndreasBackx friendly nudge. We could really use this for Apache Airflow which is dependent on the possibility of pickling

sdispater:
Fixed by #161

AndreasBackx:
@bolkedebruin sorry for leaving this in the dust. I've been getting swamped these last few weeks. Thanks for putting forth your own PR. :+1:

PR #139, opened 2017-08-26
_compat.py Use feature detection instead of version detection

cclauss:
https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection

On `encodings`, given the `return` statement, all other _falsey_ values are just as dangerous as `None`.


sdispater:
Thanks!

PR #140, opened 2017-08-26
pendulum.from_format() can now accept a list of formats

2x2xplz:
The Arrow library allows its get() function to accept a list of possible formats. That library will attempt to parse a string with each of the formats in the list until it finds one that works. My goal was to copy this functionality into pendulum.from_format(), which currently only accepts a single string and therefore can only try one format per statement. Of course, a single format string will continue to work.

Example:
`pendulum.from_format(a_date_string, ['%m/%d/%Y', '%Y-%m-%d', '%b %d %Y'])`

### DIFF
--- a/pendulum/pendulum.py
+++ b/pendulum/pendulum.py
@@ -378,7 +378,7 @@ class Pendulum(Date, datetime.datetime):
         Create a Pendulum instance from a specific format.
 
         :param fmt: The format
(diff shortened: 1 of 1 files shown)


sdispater:
Thanks for taking the time to make this PR!

Could you add some tests covering this feature?

Issue #137, opened 2017-08-23
Averages don't take into account milliseconds


ericfrederich:
Averages seem to just copy the millisecond portion instead of averaging them.

For two instances of Pendulum `p1` and `p2` I believe it should hold that `p1.average(p2) == p2.average(p1)`

This is not currently the case.  In fact, averaging two times that are only 1 second apart will yield one of the source times rather than a midpoint.

```Python
import pendulum
import time

p1 = pendulum.now()
time.sleep(1)
p2 = pendulum.now()

a1 = p1.average(p2)
a2 = p2.average(p1)

print(p1)
print(a1)
print(a2)
print(p2)
print(a1 == a2)
print(p1 == a1)
print(p2 == a2)
```
**example output:**
```
2017-08-23T12:01:32.809279-04:00
2017-08-23T12:01:32.809279-04:00
2017-08-23T12:01:33.821634-04:00
2017-08-23T12:01:33.821634-04:00
False
True
True
```

sdispater:
Yes, you are definitely right.

I just checked the code and `average()` only averages at the `seconds` level and not microseconds.

It's an easy fix so it should not be too long before it makes it to the next minor release

ericfrederich:
@sdispater, great!  Add a test too if you can.  I currently have datetime code that recursively bisects intervals.  I'd like to switch to pendulum but obviously cannot until this is fixed.

ericfrederich:
Pinging @sdispater ... I wrote a fix and some tests.

Things in 2.0 branch are refactored so it's not a clean merge.  I can merge conflicts and get 2.0 branch up to date once this is merged into master.

ericfrederich:
Look good @sdispater ?

sdispater:
@ericfrederich Yes, your PR fixed this. However I will close this once I publish the next release.

PR #142, opened 2017-08-28
Incorporate microseconds into averages

ericfrederich:
Fixes issue #137

### DIFF
--- a/pendulum/pendulum.py
+++ b/pendulum/pendulum.py
@@ -1769,7 +1769,8 @@ class Pendulum(Date, datetime.datetime):
         if dt is None:
             dt = Pendulum.now(self._tz)
 
-        return self.add(seconds=int(self.diff(dt, False).in_seconds() / 2))
+        diff = self.diff(dt, False)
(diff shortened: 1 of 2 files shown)


sdispater:
Thanks for taking the time to make this PR.

I just pushed a commit (fcec7513ebee984f0b9edc72d9e4e67f377c2dcb) on `master` that fixes how microseconds are normalized in intervals. You can check https://pendulum.eustace.io/docs/#interval to see what I mean by normalization.

```python
>>> import pendulum

>>> d1 = pendulum.create(2012, 1, 1, 1, 2, 3, 123456)
>>> d2 = pendulum.create(2011, 12, 31, 22, 2, 3)
>>> delta = d2 - d1
>>> delta.seconds
-10800
>>> delta.microseconds
-123456
```

This change makes it easier to include microseconds into averages since you no longer need to compute seconds and microseconds in `average()`, you can just use `add()` directly.

ericfrederich:
@sdispater great... the code for `average` is much less ugly now!

This pull request has been re-based.  The same tests that I wrote for the previous implementation still pass.

sdispater:
Thanks a lot!

PR #158, opened 2017-11-08
Fix instance() behaviour with pytz offsets

pR0Ps:
`pytz` has a class of timezone called a `FixedOffset`. This class only
stores an offset, not a zone name.

When calling `instance()` on a datetime that was using one of these
`FixedOffset` objects as a timezone, Pendulum would totally ignore the
offset and use the local timezone instead. This would happen even if a
`tz` argument was directly supplied to the `instance()` call.

Additionally, the fallback to a fixed offset would always use the
datetime's timezone, even it if was `None` and a `tz` argument was
supplied to the `instance()` call.

This commit makes 2 small changes:
- The `zone` property of the "effective timezone" (the datetime's
  `tzinfo` with a fallback to the passed in `tz`) is only used if it is
  non-null. This fixes using `FixedOffset` objects as timezones.
- The fallback to a fixed offset now uses the offset of the "effective
  timezone" instead of the datetime's offset. This correctly handles
  cases where the timezone on the datetime is `None` and a fallback is
  provided.

### DIFF
--- a/pendulum/pendulum.py
+++ b/pendulum/pendulum.py
@@ -205,13 +205,13 @@ class Pendulum(Date, datetime.datetime):
         # Checking for pytz/tzinfo
         if isinstance(tz, datetime.tzinfo) and not isinstance(tz, (Timezone, TimezoneInfo)):
             # pytz
-            if hasattr(tz, 'localize'):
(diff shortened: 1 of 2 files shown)


sdispater:
Nice catch! Thanks a lot for this PR :-)

Issue #171, opened 2018-01-05
Default timezone isn't utc?


leafonsword:
From [FAQ](https://pendulum.eustace.io/faq/). 
`It also removes the notion of naive datetimes: each Pendulum instance is timezone-aware and by default in UTC for ease of use.`
But from my test, seems dfault timezone is local machine's timezone, my machine's time zone is 'Asia/Shanghai'
```
>>> pendulum.now()                                                                                                                                               
<Pendulum [2018-01-05T15:51:55.975424+08:00]>       
                                                                                                             
>>> pendulum.now('Asia/Shanghai')                                                                                                                                
<Pendulum [2018-01-05T15:52:02.086741+08:00]>      
                                                                                                              
>>> pendulum.now().in_timezone('UTC')                                                                                                                            
<Pendulum [2018-01-05T07:52:27.614535+00:00]>                                                                                                                    
>>>                                                                                                                
```
So now() has same result like now('Asia/Shanghai'), not utc

SethMichaelLarson:
Can confirm in the code that `Pendulum.now()` uses `datetime.datetime.now()` not `datetime.datetime.utcnow()`.

pganssle:
@SethMichaelLarson I would certainly hope it doesn't use `datetime.datetime.utcnow()`. Even if you want a datetime in `UTC` you shouldn't use `utcnow`, you should use `datetime.datetime.now(UTC)` instead.

SethMichaelLarson:
Could you explain the difference between the two?
Pendulum uses `datetime.datetime.utcnow().replace(tzinfo=UTC)` for the UTC case.

pganssle:
I will make a PR soon to remove the use of `utcnow` from Pendulum's internals, as it is pointless. You can see my comment [here](https://github.com/crsmithdev/arrow/pull/486#issuecomment-349951436) which explains why `utcnow` is a pointless aberration.

SethMichaelLarson:
Thanks for the clarification. I'd just say that a lot of us are here to learn and that programming with dates and times (even without timezones!) is really difficult, even for people who have a good control of the language.

pganssle:
[...]

PR #172, opened 2018-01-05
Improve Pendulum.now and Pendulum.utcnow performance

pganssle:
Per my discussion in https://github.com/sdispater/pendulum/issues/171#issuecomment-355575650 and the same reasoning as https://github.com/crsmithdev/arrow/pull/486#issuecomment-349951436, I think that `pendulum` should take advantage of the fact that `datetime.now()` takes a `tzinfo` object as an argument to provide aware datetimes.

In Python 3.6 on my Arch Linux laptop. With this patch:

```python
In [1]: from pendulum import Pendulum

In [2]: %timeit Pendulum.utcnow()
The slowest run took 4.25 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 11.4 µs per loop

In [3]: %timeit Pendulum.now('UTC')
The slowest run took 6.06 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 11.3 µs per loop

In [4]: %timeit Pendulum.now()
The slowest run took 628.80 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 18.7 µs per loop
```

Before this patch:

```python
In [1]: from pendulum import Pendulum

In [2]: %timeit Pendulum.utcnow()
The slowest run took 4.17 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 11.1 µs per loop

In [3]: %timeit Pendulum.now('UTC')
100000 loops, best of 3: 11.1 µs per loop

In [4]: %timeit Pendulum.now()
The slowest run took 228.83 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 33 µs per loop
```

I guess it doesn't have a major effect on the speed of `utcnow()` for whatever reason, but it at the very least doesn't have a *negative* performance impact, and it's also the right thing to do when constructing aware datetimes.

### DIFF
--- a/pendulum/pendulum.py
+++ b/pendulum/pendulum.py
@@ -273,11 +273,11 @@ class Pendulum(Date, datetime.datetime):
             return test_instance
 
         if tz is None or tz == 'local':
-            dt = datetime.datetime.now()
+            dt = datetime.datetime.now(cls._local_timezone())
         elif tz is UTC or tz == 'UTC':
-            dt = datetime.datetime.utcnow().replace(tzinfo=UTC)
+            dt = datetime.datetime.now(UTC)
         else:
-            dt = datetime.datetime.utcnow().replace(tzinfo=UTC)
+            dt = datetime.datetime.now(UTC)
             tz = cls._safe_create_datetime_zone(tz)
             dt = tz.convert(dt)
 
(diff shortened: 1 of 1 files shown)


pganssle:
There are also several places in the tests where you use `utcnow()` when `now(UTC)` would work just fine, but I didn't want to touch those because I was not entirely sure if constructing an instance from a naive datetime was actually part of the test for some reason.

It does seem to me that a lot of the tests use `now` or `utcnow()` when they could use datetime literals instead, I'm not sure why that was done that way.

sdispater:
Thanks for that !

Regarding the use of `now()` or `utcnow()` in the tests, it's just for conciseness more than anything.

PR #161, opened 2017-11-14
Add deepcopy / pickle compatibility

bolkedebruin:
Python's doc say: "A tzinfo subclass must have an __init__() method that can
be called with no arguments". TimezoneInfo doesn't honor this
requirement. Defining __getinitargs__ is sufficient to fix copy/deepcopy as
well as pickling/unpickling.

This is https://github.com/sdispater/pendulum/pull/134 implemented

PR #174, opened 2018-01-26
Fix unpickling with undefined / empty timezone name

### LINKED ISSUE #167: Timezone is empty on Windows 10 bash shell, causing error while unpickling Pendulum objects
Hello.

This is maybe related to #161, but I'm not sure.

When I run this using the Windows 10 bash shell:
```python
import pendulum
import pendulum.version
import pickle
import sys

print("Python version:", sys.version)
print("Pendulum version:", pendulum.version.VERSION)

now = pendulum.now()
print(now, now.timezone)
pickled = pickle.dumps(now)
pickle.loads(pickled)
```

It raises an error (but it works fine with W10 Python installation):
```
Python version: 3.6.3 (default, Oct  4 2017, 02:55:45) [GCC 5.4.0 20160609]
Pendulum version: 1.3.2
2017-12-11T20:31:49.750107+01:00 <Timezone []>
  File "test.py", line 12, in <module>
    pickle.loads(pickled)
  File "/mnt/c/Users/delgan/Desktop/project/env/lib/python3.6/site-packages/pendulum/pendulum.py", line 155, in __init__
    self._tz = self._safe_create_datetime_zone(tzinfo)
  File "/mnt/c/Users/delgan/Desktop/project/env/lib/python3.6/site-packages/pendulum/pendulum.py", line 82, in _safe_create_datetime_zone
    return cls._timezone(obj)
  File "/mnt/c/Users/delgan/Desktop/project/env/lib/python3.6/site-packages/pendulum/pendulum.py", line 107, in _timezone
    return Timezone.load(zone)
  File "/mnt/c/Users/delgan/Desktop/project/env/lib/python3.6/site-packages/pendulum/tz/timezone.py", line 90, in load
    utc_transition_times) = Loader.load(name)
  File "/mnt/c/Users/delgan/Desktop/project/env/lib/python3.6/site-packages/pendulum/tz/loader.py", line 33, in load
    with tz_file(name) as f:
  File "/mnt/c/Users/delgan/Desktop/project/env/lib/python3.6/site-packages/pytzdata/__init__.py", line 32, in tz_file
    filepath = tz_path(name)
  File "/mnt/c/Users/delgan/Desktop/project/env/lib/python3.6/site-packages/pytzdata/__init__.py", line 62, in tz_path
    raise ValueError('Invalid timezone')
ValueError: Invalid timezone
```

As you can see, the timezone is empty, which causes an exception equivalent to `pendulum.now(tz='')`. Yet, the `now()` worked fine because the time contains `+01:00`.

Do you have any idea where the bug might come from?

Delgan:
This PR is intended to fix #167.

Basically, the issue was that Pendulum defaults to `""` for the local timezone name [if it can't be determined properly](https://github.com/sdispater/pendulum/blob/74350b3532c96c2cdace651833e2524fc54c97e8/pendulum/tz/local_timezone.py#L161-L167), but the `tz` attribute of `Pendulum()` does not accept empty string. This was causing an error while trying to re-create a pickled instance of `Pendulum`.

To fix that, I made use of #161. Now that `TimezoneInfo` objects are picklable, we can directly use them to serialize any instance of Pendulum, rather than their string representation.

I also had to fix a small bug  in the `__getinitargs__()` method of the `FixedTimezone` class, as [it should return a tuple](https://docs.python.org/2/library/pickle.html#object.__getinitargs__) but it did not.

Any `Pendulum` instance with an empty timezone name should now be correctly unpickable.

As to why the timezone name could not be determined, it seems to be a bug with Windows 10 bash shell. Here is a discussion of another project facing the same issue: https://github.com/HowardHinnant/date/issues/102 
There does not seem to be any way to know the timezone's name. However, it is recommended to run `sudo dpkg-reconfigure tzdata` to fix it, which I did and it works fine now.

### DIFF
--- a/pendulum/pendulum.py
+++ b/pendulum/pendulum.py
@@ -1980,16 +1980,10 @@ class Pendulum(Date, datetime.datetime):
         return(self, )
 
     def _getstate(self, protocol=3):
-        tz = self.timezone_name
-
-        # Fix for fixed timezones not being properly unpickled
-        if isinstance(self.tz, FixedTimezone):
-            tz = self.offset_hours
-
         return (
             self.year, self.month, self.day,
             self.hour, self.minute, self.second, self.microsecond,
-            tz,
+            self.tzinfo,
             self.fold
         )
 
--- a/pendulum/tz/timezone.py
+++ b/pendulum/tz/timezone.py
@@ -522,7 +522,7 @@ class FixedTimezone(Timezone):
         return (dt + self._tzinfo.adjusted_offset).replace(tzinfo=self._tzinfo)
 
     def __getinitargs__(self):
-        return self._offset
+        return (self._offset, )
 
 
 class _UTC(FixedTimezone):
--- a/tests/pendulum_tests/test_behavior.py
+++ b/tests/pendulum_tests/test_behavior.py
@@ -6,6 +6,7 @@ from copy import deepcopy
 from datetime import datetime, date, time, timedelta
 from pendulum import Pendulum, timezone
 from pendulum.tz.timezone import Timezone
+from pendulum.tz.loader import Loader
 from .. import AbstractTestCase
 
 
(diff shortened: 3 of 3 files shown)


sdispater:
Thanks a lot for this and the detailed explanation !

One thing though: Could you base your PR on the `1.x` branch?

The `master` branch will soon be holding the work towards the `2.0` release so it will have some breaking changes.

Delgan:
Done. 👍

sdispater:
Thanks !

PR #177, opened 2018-02-10
Fix exceptions unpickling UTC objects

pR0Ps:
Without this commit, the `_UTC` object inherits the `TimezoneInfo`'s `__getinitargs__` which tries to initialize the object with 5 extra args, causing a `TypeError`.

(This error is new in v1.4.1)


pR0Ps:
Fixed in 374d1da7c384797ef0d215418bb331aa30ff09d1

PR #195, opened 2018-04-18
utcnow removed

jgirardet:


Julian:
Why was this removed?

sdispater:
@Julian Because there is no need for it since you can just do `pendulum.now('UTC')`

Julian:
Yes but it was an existing API -- I'm not saying the other way is
difficult, but removing a public API (even when you bump major versions)
still creates API churn on everyone downstream to update their code (all
our unit tests started failing, and I had to locate this commit after
diagnosing what was up).

Fair enough though if that was all here.

Julian:
This isn't a semver thing. Semver tells you what to do *after* you decided
what change to make, not what changes to make.

It's not a big deal, but my intention here was to say "this change seems
spurious".

And we do pin versions, which is why nothing more than CI broke, our CI
intentionally runs on both the pinned and unpinned versions.

bmjjr:
FYI, ran into an upgrade issue where the lack of this equivalent API for utcnow is fairly painful. The existing code uses `datetime.datetime.utcnow` without calling it, as a "DATETIME_FACTORY".  So I can't just use `pendulum.now('UTC')` as an instanciated object, and seems like just using `pendulum.now` is not the right answer given `UTC` is not specified and so it is returning `Timezone('Asia/Shanghai')` when called instead of Timezone('UTC').

Issue #206, opened 2018-05-11
Period has no __hash__ or __eq__ which makes the behavior of set() highly confusing


mshuffett:
Any set of periods that has the same duration are seen as equal. This is highly confusing and decreases the usefulness of Period.

See the below code example:
```python
import pendulum
period1 = pendulum.DateTime(2018, 1, 3) - pendulum.DateTime(2018, 1, 2)
period2 = pendulum.DateTime(2018, 1, 2) - pendulum.DateTime(2018, 1, 1)
s = {period1, period2}
assert len(s) == 1  # 2 would be the expected result
```

mshuffett:
Closed with #207

PR #207, opened 2018-05-11
Added __eq__ and __hash__ to Period

mshuffett:
This pull request solves #206 by adding __eq__ and __hash__ to Period.

### DIFF
--- a/pendulum/period.py
+++ b/pendulum/period.py
@@ -346,3 +346,9 @@ class Period(Duration):
 
     def __reduce_ex__(self, protocol):
         return self.__class__, self._getstate(protocol)
+
+    def __hash__(self):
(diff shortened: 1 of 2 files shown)


Marlysson:
Why does use the hash method if you can just use the __eq__ method comparing some attributes?

sdispater:
@Marlysson Implementing the __hash__ method makes it possible to use Period obkects has key in a dictionnary for example.

sdispater:
@mshuffett Thanks!

Marlysson:
@sdispater But if the Period object doesn't wasn't used in a dictionary , the hash method, still would be necessary?

sdispater:
@Marlysson In this case, no. Equality (`==`) is just checked with the `__eq__` method.

Issue #215, opened 2018-05-16
PendulumException should derive from Exception instead of BaseException


jgaffiot:
Hi,

as is stated in the [official documentation](https://docs.python.org/3/library/exceptions.html):
> programmers are encouraged to derive new exceptions from the Exception class or one of its subclasses, and not from BaseException

The rational is discussed in [PEP352](https://www.python.org/dev/peps/pep-0352/) or this [StackOverflow discussion](https://stackoverflow.com/questions/27995057/why-is-it-recommended-to-derive-from-exception-instead-of-baseexception-class-in).

Moreover, I think most programmers (at least me!) expect that all errors are caught in a `except Exception` block, which is currently not the case, you have to:

    except (Exception, pendulum.exceptions.PendulumException):

to be sure to catch everything but the system exit, keyboard interrupt...
Perhaps it's still time for v2 ?

Best regards and thanks for the amazing library!

PR #217, opened 2018-05-17
Make "PendulumException" a subclass of "Exception"

Delgan:
I'm a little bit ashamed of such a small PR... 😁 

Reference: #215


sdispater:
Thanks!

And there is not such things as a small PR as long as it's useful :-)

PR #229, opened 2018-06-07
Fix localization section in README

Senpos:
- wrong URL for `locales/en/custom.py`
- wrong command for generating locale with `clock`


sdispater:
Thanks!

PR #236, opened 2018-06-15
Fix import error when using `pyinstaller`

NateScarlet:
File discover is unavaliable when run with `pyinstaller`.
I added a `index.py` to save avaliable locale.
See https://github.com/pyinstaller/pyinstaller/issues/3528


RudolfCardinal:
As an alternative method, see suggested PyInstaller hook for Pendulum at https://github.com/pyinstaller/pyinstaller/issues/3528 (which includes its locale data).

PR #237, opened 2018-06-15
Fixed IANA tz database url

JHoloboski:
Really simple docs fix, just found this while going through them. The url currently in instantiation for the IANA time zone database is invalid, changed it to the right url.


sdispater:
Thanks!

mwoyuan:
This patch didn't seem to go live: https://pendulum.eustace.io/docs/#instantiation the link still goes to the old wrong one.

PR #243, opened 2018-06-24
Improve regexes and reduce string manipulation

lint-ai:


pendulum/tz/local_timezone.py
```diff
@@ -177,9 +177,7 @@ def _get_unix_timezone(_root='/'):  # type: (str) -> Timezone
     # OpenSUSE has a TIMEZONE setting in /etc/sysconfig/clock and
     # Gentoo has a TIMEZONE setting in /etc/conf.d/clock
     # We look through these files for a timezone:
-    zone_re = re.compile('\s*ZONE\s*=\s*\"')
-    timezone_re = re.compile('\s*TIMEZONE\s*=\s*\"')
-    end_re = re.compile('\"')
+    timezone_re = r'\s*(TIME)?ZONE\s*=\s*"(?P<zone_path>.*?)"'
```
sdispater: It should be a regex object. Right now it breaks since `match()` is called on a string.

lint-ai: @sdispater fixed

Issue #210, opened 2018-05-15
Inconsistency with 'parse("13:00")' depending if Pendulum is built with C extensions or not


Delgan:
Hi.

I made some tests with two different installations of Pendulum v2.0.1

_First one (Windows)_
```python
>>> import pendulum
>>> pendulum.parse("13:00")
DateTime(2018, 5, 15, 13, 0, 0, tzinfo=Timezone('UTC'))
```

_Second one (Linux)_
```python
>>> import pendulum
>>> pendulum.parse("13:00")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/c/Users/delgan/env/lib/python3.6/site-packages/pendulum/parser.py", line 20, in parse
    return _parse(text, **options)
  File "/mnt/c/Users/delgan/env/lib/python3.6/site-packages/pendulum/parser.py", line 36, in _parse
    parsed = base_parse(text, **options)
  File "/mnt/c/Users/delgan/env/lib/python3.6/site-packages/pendulum/parsing/__init__.py", line 64, in parse
    return _normalize(_parse(text, **_options), **_options)
  File "/mnt/c/Users/delgan/env/lib/python3.6/site-packages/pendulum/parsing/__init__.py", line 116, in _parse
    raise ParserError('Unable to parse string [{}]'.format(text))
pendulum.parsing.exceptions.ParserError: Unable to parse string [13:00]
```

I guess this is somehow related to C extensions, because `from pendulum.parsing._iso8601 import parse_iso8601` raises a `ModuleNotFound` exception on my Windows, but not on Linux.

On the other hand, `from pendulum._extensions._helpers import local_time` works fine on both Linux and Windows, so it confuses me a little bit... 😕

sdispater:
This should be fixed in the latest release: `2.0.2`

Delgan:
@sdispater Thanks for the fix!

However, there is still a problem with `pendulum.parse("13")` which raises an exception, but shouldn't.

PR #251, opened 2018-07-07
Fix exact parsing of hour-only strings

Delgan:
Hi.

I commented on [#210](https://github.com/sdispater/pendulum/issues/210#issuecomment-393293238) a few weeks ago about a small bug while parsing:

> However, there is still a problem with `pendulum.parse("13")` which raises an exception, but shouldn't.

I copied your previous fix and extended it so it handles parsing of times without minutes nor seconds.

Also, I added some unit tests to prevent regression on edge cases like this, and I split the `test_parse_exact()` function for better readability. 👍

### DIFF
--- a/pendulum/parsing/__init__.py
+++ b/pendulum/parsing/__init__.py
@@ -28,7 +28,7 @@ COMMON = re.compile(
     # Time (optional)
     '(?P<time>'
     '    (?P<timesep>\ )?'  # Separator (space)
-    '    (?P<hour>\d{1,2}):(?P<minute>\d{1,2})?(?::(?P<second>\d{1,2}))?'  # HH:mm:ss (optional mm and ss)
+    '    (?P<hour>\d{1,2})(?::(?P<minute>\d{1,2})(?::(?P<second>\d{1,2}))?)?'  # HH:mm:ss (optional mm and ss)
     # Subsecond part (optional)
     '    (?P<subsecondsection>'
     '        (?:[.|,])'  # Subsecond separator (optional)
@@ -169,7 +169,10 @@ def _parse_common(text, **options):
     # Grabbing hh:mm:ss
     hour = int(m.group('hour'))
 
-    minute = int(m.group('minute'))
+    if m.group('minute'):
+        minute = int(m.group('minute'))
+    else:
+        minute = 0
 
     if m.group('second'):
         second = int(m.group('second'))
--- a/tests/test_parsing.py
+++ b/tests/test_parsing.py
@@ -43,7 +43,14 @@ def test_parse_with_timezone():
     assert 7200 == dt.offset
 
 
-def test_parse_exact():
+def test_parse_exact_datetime():
+    text = '2016-10-16T12:34:56.123456'
+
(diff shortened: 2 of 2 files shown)


sdispater:
"13" is too ambiguous to be parsed reliably so I don't think this is a good idea. In this particular case it's better to use the `from_format()` with the `HH` token.

Delgan:
@sdispater This is true, I did not even think to this, I just tried to mimic the pre-2.0 behavior (and to match comments in the code).

After a short thought, I would say this is mainly the user's concern, as long as `Pendulum` is clear that two-digits strings are parsed as hours, he can use `from_format()` if he needs another disambiguation.

The parallel between extended/reduced formats of dates and times would be quite intuitive in my opinion:

    2018-07-08 -> 2018-07 -> 2018
    23:55:12 -> 23:55 -> 23

And there will always be problems of ambiguity, like is `1933` parsed as `YYYY` or `HHmm`? Is `201104` equivalent to `YYYYMM` or `HHmmss`?

That's just my point of view theorically. In practice I don't care much. 😉 

However, the main goal of this PR was also to fix inconsistencies between `Pendulum` builts with or without C extensions. Because, in one case `parse("13")` is perfectly valid while it raises an exception in the other case. You can see this in [this Travis build](https://travis-ci.org/Delgan/pendulum/builds/401183406): jobs with `PENDULUM_EXTENSIONS=1` failed while [my initial commit](https://github.com/Delgan/pendulum/commit/efde00d2ea3ca496a7ff7c23463add953543a016) just added the corresponding test case (although I can't explain why my fix preventing C-based builts to fail only modifies Python code).
So, there is still a problem somewhere. 😕

Delgan:
Hey @sdispater.

There is still a remaining inconsistency, I don't know if another issue references this problem:

```python
# Pendulum built with "PENDULUM_EXTENSIONS=0"
>>> pendulum.parse("13")
DateTime(2019, 8, 8, 13, 0, 0, tzinfo=Timezone('UTC'))
```

```python
# Pendulum built with "PENDULUM_EXTENSIONS=1"
>>> pendulum.parse("13")
  File "/home/delgan/env/lib/python3.7/site-packages/pendulum/parser.py", line 20, in parse
    return _parse(text, **options)
  File "/home/delgan/env/lib/python3.7/site-packages/pendulum/parser.py", line 36, in _parse
    parsed = base_parse(text, **options)
  File "/home/delgan/env/lib/python3.7/site-packages/pendulum/parsing/__init__.py", line 70, in parse
    return _normalize(_parse(text, **_options), **_options)
  File "/home/delgan/env/lib/python3.7/site-packages/pendulum/parsing/__init__.py", line 124, in _parse
    raise ParserError("Unable to parse string [{}]".format(text))
pendulum.parsing.exceptions.ParserError: Unable to parse string [13]
```

sdispater:
@Delgan Yes, you are right. The pure Python implementation should also fail in this case

Delgan:
Fine, I will open a issue referencing this just for it to be logged explicitly. :+1:

PR #254, opened 2018-07-21
Corrects a bug on closest/farthest and extends functionality to take *tuple

oddProton:
The closest/farthest functions didn't actually consider the distance between the current instance and the two parameters. On master, the 'smaller'/'bigger' was returned regardless of the actual distance between the instance and the parameters.

``` Python
>>> a = pendulum.now()
>>> b = pendulum.now() + pendulum.duration(hours=10)
>>> c = pendulum.now() + pendulum.duration(hours=-15)
>>> a.closest(b,c) == c
True
>>> # however, the closest is actually b
>>> # similarly, using farthest, should return C but instead returns b :
>>> a.farthest(b,c) == b
True
```

In addition, I have extended the functionality to receive unpacked iterables, through * operator while keeping the initial signature and backwards compatibility.

The results of the test:

>  py27: commands succeeded
ERROR:   py35: InterpreterNotFound: python3.5
  py36: commands succeeded
  pypy: commands succeeded
ERROR:   pypy3: InterpreterNotFound: pypy3

### DIFF
--- a/pendulum/datetime.py
+++ b/pendulum/datetime.py
@@ -452,36 +452,43 @@ class DateTime(datetime.datetime, Date):
         )
 
     # Comparisons
-    def closest(self, dt1, dt2):
-        """
-        Get the closest date from the instance.
-
-        :type dt1: DateTime or datetime
-        :type dt2: DateTime or datetime
-
-        :rtype: DateTime
-        """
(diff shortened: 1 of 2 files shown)


sdispater:
Thanks!

PR #268, opened 2018-08-24
Corrected 'z' timezone parsing in from_format to allow underscores

cordery:
Ex: America/New_York.  Expanded tests to ensure parsing of all symbols found in pendulum.timezones.


sdispater:
Thanks!

PR #280, opened 2018-09-14
Add Norwegian locales

maedox:
Norway has two locales, bokmål (nb) and Nynorsk (nn). This adds both with tests.


maedox:
That went well... Tests run successfully locally on Python 3.7 at least.
I suppose the problem is elsewhere as it fails on poetry install.

maedox:
Is anyone there?

sdispater:
Sorry for the delay.

Could you rebase your PR so that we make sure that the tests pass?

maedox:
Hi. I'm back! Sorry, I forgot all about this until I needed a NO locale again.
I have rebased and it still fails.

sdispater:
Thanks!

PR #285, opened 2018-10-09
Added support for parsing padded 2-digit days of the month with 'from…

### LINKED ISSUE #269: Space-padded days in from_format
First off, thanks for this great library – I’m basically using it in all of my projects. Now I’ve run into a format oddity that seems to be more common than I thought: space-padded days.

This is used by `ls`, OpenSSL, and apparently syslog.

For example what I needed to parse today is `Apr  2 12:00:00 2020 GMT` (note the two spaces before the first 2) which requires a different from_format format than say `Apr 12 12:00:00 2020 GMT`. 

It looks like datetime's strptime [handles this](https://twitter.com/moreati/status/969329950925967361) out of the box.  True to pendulum's principles I would of course prefer an explicit way of saying “space-padded day” instead of guessing.

tersers:
…_format'

Fixes #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fixed `from_format()` not recognizing input strings when the specified pattern had escaped elements.
 - Fixed missing `x` token for string formatting.
+- Added support for parsing padded 2-digit days of the month with `from_format()`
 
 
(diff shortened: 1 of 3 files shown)


sdispater:
Thanks!

PR #291, opened 2018-10-25
Incorrect age calculation for future dates

### LINKED ISSUE #264: Incorrect `age` calculation for future dates.
```python
import pendulum
birthday = pendulum.parse('9999-12-31')
birthday.age < 16
```
```
Out[5]: False
```

pendulum==2.0.3
python3.5.4

The problem is that underlying `pendulum.Date.diff` returns absolute value in years. While it looks it should be non absolute value.

```
@property
def age(self):
    return self.date().diff(self.now(self.tz).date()).in_years()
```

So I suppose that passing `abs=False` parameter to the `pendulum.Date.diff` would fix the issue.

aleksey-kutepov:
Fixes #264

### DIFF
--- a/pendulum/date.py
+++ b/pendulum/date.py
@@ -80,7 +80,7 @@ class Date(FormattableMixing, date):
 
     @property
     def age(self):
-        return self.diff().in_years()
+        return self.diff(abs=False).in_years()
 
     @property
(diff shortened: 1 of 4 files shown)


sdispater:
Thanks!

Issue #293, opened 2018-10-26
is_utc() returns False when parsing from a UTC date


joaonc:
This date is UTC: `'2018-06-21T09:30:00+00:00'`, however, the `is_utc()` method returns `False`.

```
>>> pendulum.parse('2018-06-21T09:30:00+00:00').timezone
Timezone('+00:00')

>>> pendulum.parse('2018-06-21T09:30:00+00:00').is_utc()
False
```

The current implementation is:
```
    def is_utc(self):
        return self.timezone_name == 'UTC'
```
Maybe it can be something like:
```
    def is_utc(self):
        return self.timezone.offset == 0
```

redlickigrzegorz:
I created a pull request #295 for resolving this issue. I hope it resolves the problem. I would like to ask for a review also.

joaonc:
@darrenleeweber Are you saying/asking something or simply demonstrating cases where it returns `False` when should be returning `True`?  (ie, reproing the bug)

darrenleeweber:
Yes, reproducing the behavior in 1.5.1 vs 2.x is important for me (and possibly others landing here).

darrenleeweber:
#295 seems to help with the logic in the `is_utc()` method, but that doesn't entirely resolve some concerns about the `pendulum.parse()` method and how it should behave, esp. when given an explicit timezone to apply.

joaonc:
My take is that the scope of this issue is the `is_utc()` method which, regardless of how the DateTime object was created, if the timzone has an offset of 0 (no offset, aka UTC), then `is_utc()` should return `True`, which hopefully now it does.

PR #295, opened 2018-10-27
is_utc() returns False when parsing from a UTC date

redlickigrzegorz:
It is a pull request for resolving the problem of `is_utc()` method -> #293 
I created additional two assertions and prepare the code which resolves the problem without the breaking anything else, I hope.
Furthermore, I think that this method should return `True` for `"GMT"` timezone because there is no time difference between Greenwich Mean Time and Coordinated Universal Time.


redlickigrzegorz:

sdispater:
Thanks for your contribution!

Issue #282, opened 2018-09-21
int_timestamp returns different values than timestamp()


Freewolny:
```
>>> pendulum.datetime(2018, 10, 28, 2, 20, tz = 'Europe/Warsaw', dst_rule = pendulum.PRE_TRANSITION).timestamp()
1540686000.0
>>> pendulum.datetime(2018, 10, 28, 2, 20, tz = 'Europe/Warsaw', dst_rule = pendulum.POST_TRANSITION).timestamp()
1540689600.0
>>> pendulum.datetime(2018, 10, 28, 2, 20, tz = 'Europe/Warsaw', dst_rule = pendulum.PRE_TRANSITION).int_timestamp
1540686000
>>> pendulum.datetime(2018, 10, 28, 2, 20, tz = 'Europe/Warsaw', dst_rule = pendulum.POST_TRANSITION).int_timestamp
1540686000
```

The last result should be `1540689600`, not `1540686000`. Transition isn't respected here.

redlickigrzegorz:
I created a pull request #296 for resolving this issue. I hope it resolves the problem. I would like to ask for a review also.

PR #296, opened 2018-10-27
int_timestamp returns different values than timestamp()

redlickigrzegorz:
It is a pull request for resolving the problem of `int_timestamp()` property method -> #282 
I prepared the solution which is identical to `timestamp()` method without interference to the current behavior of getting timestamp as an integer.


sdispater:
Looks good to me 👍 

Thanks!

Issue #272, opened 2018-09-07
`is_birthday` should be `is_anniversary`


jamesmyatt:
`is_birthday` is a bit specific; `is_anniversary` is more correct. Maybe it can just be added as an alias to keep backwards compatbility.

redlickigrzegorz:
I created a pull request #298 for implementing the alias. I hope it is exactly what is needed. I would like to ask for a review also. I have only a problem with tests, but it can be a more global problem because I see exactly the same output on my local machine, even on the master branch. I will try to figure out something as soon as possible.

PR #298, opened 2018-10-28
`is_birthday` should be `is_anniversary`

redlickigrzegorz:
It is a pull request for implementing an alias for `is_birthday()` method -> #272

### DIFF
--- a/pendulum/date.py
+++ b/pendulum/date.py
@@ -196,7 +196,7 @@ class Date(FormattableMixing, date):
         """
         return self == dt
 
-    def is_birthday(self, dt=None):
+    def is_anniversary(self, dt=None):
         """
         Check if its the birthday.
 
@@ -214,7 +214,7 @@ class Date(FormattableMixing, date):
     # the additional method for checking if today is the anniversary day
     # the alias is provided to start using a new name and keep the backward compatibility
     # the old name can be completely replaced with the new in one of the future versions
-    is_anniversary = is_birthday
+    is_birthday = is_anniversary
 
     # ADDITIONS AND SUBSTRACTIONS
 
--- a/pendulum/date.py
+++ b/pendulum/date.py
@@ -198,7 +198,7 @@ class Date(FormattableMixing, date):
 
     def is_anniversary(self, dt=None):
         """
-        Check if its the birthday.
+        Check if its the anniversary.
 
         Compares the date/month values of the two dates.
 
--- a/pendulum/date.py
+++ b/pendulum/date.py
@@ -207,9 +207,14 @@ class Date(FormattableMixing, date):
         if dt is None:
             dt = Date.today()
 
-        instance = dt1 = self.__class__(dt.year, dt.month, dt.day)
+        instance = self.__class__(dt.year, dt.month, dt.day)
 
         return (self.month, self.day) == (instance.month, instance.day)
+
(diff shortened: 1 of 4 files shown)


pendulum/date.py
```diff
@@ -207,10 +207,15 @@ def is_birthday(self, dt=None):
...
 
-        instance = dt1 = self.__class__(dt.year, dt.month, dt.day)
+        instance = self.__class__(dt.year, dt.month, dt.day)
 
         return (self.month, self.day) == (instance.month, instance.day)
 
+    # the additional method for checking if today is the anniversary day
+    # the alias is provided to start using a new name and keep the backward compatibility
+    # the old name can be completely replaced with the new in one of the future versions
+    is_anniversary = is_birthday
```
jamesmyatt: I would make the alias the other way round, i.e. `is_anniversary` is the original method with `is_birthday` as the copy

tests/date/test_comparison.py
```diff
@@ -137,20 +137,20 @@ def test_less_than_or_equal_false():
     assert not d1 <= d3
 
 
-def test_is_birthday():
+def test_is_anniversary():
```
jamesmyatt: You won't have any tests on `is_birthday` any more if you do this

pendulum/date.py
redlickigrzegorz: Good point, I will change it as you proposed 👍

tests/date/test_comparison.py
redlickigrzegorz: I think that if these two methods are equal, I don't need additional tests. But if you think differently, I can duplicate them or write the additional ones. What is your opinion?

redlickigrzegorz: I duplicated the tests for `is_birthday` to avoid changes provides by mistake.

jamesmyatt: As far as the unit tests are concerned, it doesn't know that they're equal. So I think your options are:

- duplicate the tests (as you've done)
- test attributes are equal (this may not be a good idea, since you may want to change one or the other later)
- parameterize tests to run the same tests on both methods.

pendulum/datetime.py
```diff
@@ -571,7 +571,7 @@ def is_same_day(self, dt):
 
         return self.to_date_string() == dt.to_date_string()
 
-    def is_birthday(self, dt=None):
+    def is_anniversary(self, dt=None):
```
jamesmyatt: docstring needs updating

tests/datetime/test_comparison.py
```diff
@@ -237,21 +237,26 @@ def test_less_than_or_equal_with_timezone_false():
     assert not d1 <= d3
 
 
-def test_is_birthday():
+def test_is_anniversary():
```
jamesmyatt: I would have parametrized this, or made 2 independent copies

pendulum/datetime.py
redlickigrzegorz: You are definitely right 👍 I will update, thanks!

tests/datetime/test_comparison.py
redlickigrzegorz: Two independent copies sound reasonable for me 👍

redlickigrzegorz:

jamesmyatt:
Good enough for me

Issue #257, opened 2018-07-27
Formatter token list in documentation is incomplete


ZaneHintzman:
I discovered by accident that the `format()` method accepts a single `Y` as a valid token, but this is not in the list of valid tokens in the Formatter documentation:

https://pendulum.eustace.io/docs/#formatter

I'm using pendulum version 2.0.2.

redlickigrzegorz:
I created a pull request #299 for adding the additionally supported token. I hope it is exactly what is needed. I would like to ask for a review also. I have only a problem with tests, but it can be a more global problem because I see exactly the same output on my local machine, even on the master branch. I will try to figure out something as soon as possible.

Secrus:
Related PR was merged, so I assume the issue is resolved now. Closing.

PR #299, opened 2018-10-28
Formatter token list in documentation is incomplete

redlickigrzegorz:
It is a pull request for adding the additional token which is supported for the year's string formatting -> #257


redlickigrzegorz:

Issue #288, opened 2018-10-18
Make Duration().in_words() return "0 milliseconds"


rominf:
When I create zero length `Duration` I expect that `in_words()` will return "0 milliseconds" but not an empty string:
```
from pendulum import Duration 
d = Duration()
d.in_words()
# ''
```

redlickigrzegorz:
I created a pull request #301 for implementing the new `in_words()` representation. I hope it is exactly what is needed. I would like to ask for a review also.

PR #301, opened 2018-10-30
Make Duration().in_words() return "0 milliseconds"

redlickigrzegorz:
It is a pull request for implementing the new `in_words()` representation for the duration or period object -> #288


redlickigrzegorz:

sdispater:
Thanks!

PR #315, opened 2018-11-24
py37 for tox.ini

sheerluck:


sdispater:
The `tox.ini` file has already been updated on `master`.

PR #320, opened 2018-11-30
Update typing comments and add py.typed for PEP-561 compliance

bryanforbes:


liliserf:
This looks great! Any estimate on when it might make it in?

txomon:
Hello people, is there any chance we can get this forward?

sdispater:
Thanks!

pendulum/parser.py
```diff
@@ -13,7 +14,7 @@
 from .tz import UTC
 
 
-def parse(text, **options):
+def parse(text, **options):  # type: (str, **typing.Any) -> str
```
elemoine: Is the return type really `str`? I am not familiar with the pendulum code base, but this looks suspicious to me.

schot: I believe the correct return type is `typing.Union[pendulum.Date, pendulum.Time, pendulum.Duration]`.

PR #321, opened 2018-11-30
Handle transition.previous being None

bryanforbes:
This PR is pretty simplistic, but I found these issues when I was doing the typing updates. `pendulum.tz.zoneinfo.Transition.previous` can be `None` and `mypy` was flagging it, so I added guards for that condition and a test for `America/Chicago` for `1883-11-18`.


sdispater:
Thanks for your contribution!

PR #350, opened 2019-03-03
Fix case where transition can be None

jacebrowning:
When calling `.subtract(days=days)` with the maximum value ("now" - `datetime.min`), the following exception is raised:

```
>       if transition.is_ambiguous(sec):
E       AttributeError: 'NoneType' object has no attribute 'is_ambiguous'
```

### DIFF
--- a/pendulum/tz/timezone.py
+++ b/pendulum/tz/timezone.py
@@ -108,25 +108,26 @@ class Timezone(tzinfo):
             else:
                 transition = transition.previous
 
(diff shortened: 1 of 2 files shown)


jacebrowning:
@sdispater Please let me know if there is a cleaner way to solve this.

FYI: The `InvalidPosixSpec` test seems to be failing on `master`.

sdispater:
Thanks for your contribution!

PR #353, opened 2019-03-09
Add timespec argument to to_iso8601_string() function

### LINKED ISSUE #347: Add `timespec` argument to `to_iso8601_string()`
In Python 3.6, the `timespec` argument was added to [`datetime.isoformat(...)`](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) and for good reason: depending on whether or not the `datetime` object has milliseconds, the `isoformat()` method returns a different structured string.

Pendulum's `to_iso8601_string()` has the same issue.

Here's how I resolved it:
```
    iso_8601 = dt.to_iso8601_string()
    if len(iso_8601) > 25:
        # Contains microseconds and milliseconds. Remove.
        iso_8601 = iso_8601[:19] + iso_8601[26:]

    return iso_8601
```
However, I think this type of workaround (hack, really) should not be needed and the option to return only to a certain precision (hours, minutes, seconds, milliseconds and microseconds) should be supported in pendulum, as it is in datetime.

Note: this issue is similar to #191 but I think things changed since then, given the support of this feature by Python AND the fact that there are real world scenarios where it's needed.

Itay4:
Fixes https://github.com/sdispater/pendulum/issues/347

### DIFF
--- a/pendulum/datetime.py
+++ b/pendulum/datetime.py
@@ -45,7 +45,7 @@ class DateTime(datetime.datetime, Date):
     _FORMATS = {
         "atom": ATOM,
         "cookie": COOKIE,
-        "iso8601": lambda dt: dt.isoformat(),
+        "iso8601": lambda dt, timespec: dt.isoformat(timespec=timespec),
         "rfc822": RFC822,
         "rfc850": RFC850,
         "rfc1036": RFC1036,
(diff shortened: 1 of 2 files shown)


Itay4:
currently python 2.7 build fails, as datetime for it does not support `timespec` argument
how would you like me to handle that?
should i check the version and split code accordingly ?

joaonc:
I would love to see this PR merged, however doing it would mean that the behavior would be different for Python 3.6+ and lower versions (note: not just python 2.7, as this argument was introduced in 3.6).

On the good side, the 3.5- users wouldn't see any change, they just wouldn't be able to use the new feature.

The options are:
1. Leave as-is and the new parameter would only be available in 3.6+  (easiest).
This would probably mean check for version and split code.
2. Implement the whole `timespec` functionality in pendulum and make it available for everyone.
This would open a whole new can of worms, probably (haven't checked the implementation in the `datetime` package.

My suggestion is going with option 1 and make a note of it in the documentation.
In 3.5-, take the argument but do nothing with it.

Itay4:
Can someone assist and explain why Travis build fails ?

joaonc:
@Itay4 looks like it was the _linting_ step that failed, meaning the warning that _black_ is issuing [here](https://travis-ci.org/sdispater/pendulum/jobs/507265287#L499) is causing the build to fail. Something with code style, is my guess.

Secrus:
Outdated

PR #357, opened 2019-03-11
Fix raising exception for outside scope with nth_of_month()

altendky:


altendky:
Failures all appear to be caused by #358.

altendky:
Oops, broken on master anyways.  I'll update again after that gets fixed and check this.

sdispater:
Thanks for your contribution!

PR #368, opened 2019-05-04
small typo/word choice/punctuation fixes on reading entire document

andrewschultz:
Let me know if there are too many changes in here and I need to break this pull request up. I tried to be conservative and avoid personal-style suggestions.

I hope this pull request is a nice small way to give back for all the Pendulum library has helped me with, because I know how distracting proofreading can be. (I've only had it installed for a week, too.)


AEHamrick:
I'd love to see this approved, even minor typos are a bit jarring in an otherwise very polished set of docs :)

sdispater:
Thanks!

PR #372, opened 2019-05-13
Reject text after matching string in from_format()

### LINKED ISSUE #371: Extra text allowed after matching time with from_format()
`re.match()` doesn't anchor to the end of the string thus allowing extra text without an error.

https://github.com/sdispater/pendulum/blob/94d28b0d3cb524ae02361bd1ed7ea03e2e655e4e/pendulum/formatting/formatter.py#L406-L407

altendky:
Fixes #371


sdispater:
Thanks!

PR #374, opened 2019-05-13
Correct regex A-z typo

### LINKED ISSUE #373: Typo in formatting.formatter._MATCH_TIMEZONE (A-z -> A-Z)
https://github.com/sdispater/pendulum/blob/94d28b0d3cb524ae02361bd1ed7ea03e2e655e4e/pendulum/formatting/formatter.py#L36

Should be `A-Z` not `A-z`.  The latter accepts a few characters such a `[`.

altendky:
Fixes #373.

As is, the two cases are not distinguishable because of the error reported in #371.  After #371 is fixed this test can identify that the proper exception was raised.

WIP for:
- [x] Needs #372 to test properly


altendky:
Oops, broken on master anyways.  I'll update again after that gets fixed and check this.

sdispater:
Thanks for your contribution!

PR #380, opened 2019-06-17
Fix period raising exceptions in __eq__()

gnattishness:
`Period` can raise an AttributeError when being compared with other types.

e.g. 
``` pycon
>>> import pendulum
>>> p = pendulum.now() - pendulum.yesterday()
>>> p == None
AttributeError: 'NoneType' object has no attribute 'start'
>>> # should return False
```

This fix retains the same duck-typing capabilities, but using `isinstance()` is another option.


Secrus:
Outdated

PR #386, opened 2019-07-02
Replaced windows incompatible char

### LINKED ISSUE #381: 'utf-8' codec can't decode byte 0xe9 in position 185: invalid continuation byte
Current version does not install on Windows. Has been confirmed several times by multiple people. See here: https://github.com/pypa/pip/issues/6566

Vegemash:
When pip installing under windows the char 0xe9 doesn't decode. Shame it's the e in the authors name.
Fixes #381

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015 Sébastien Eustace
+Copyright (c) 2015 Sebastien Eustace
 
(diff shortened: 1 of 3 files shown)


sdispater:
There is no longer need for this PR since it has been fixed upstream in Poetry.

PR #393, opened 2019-08-08
Add italian [it] locale.

michele-comitini:


michele-comitini:
@sdispater
> You should add tests for the new locale and you have to fix the linting issues before I can merge this
done! as unfortunate it may sound,  today poetry is failing before actual tests can be run under _CI_
_**py.test**_ run locally passes all tests.

sdispater:
Thanks!

PR #403, opened 2019-08-31
added weekdays to duration helper

jenn0pal:
helper for adding and subtracting business days (monday - friday)

### DIFF
--- a/pendulum/datetime.py
+++ b/pendulum/datetime.py
@@ -592,17 +592,18 @@ class DateTime(datetime.datetime, Date):
         years=0,
         months=0,
         weeks=0,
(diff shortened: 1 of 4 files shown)


jenn0pal:
how come this fails?

sdispater:
Thanks for you contribution!

However, week days are tricky to get right and depend on the country and/or region so I don't think adding it to Pendulum is the right call.

jenn0pal:
Hi @sdispater 

This is similar to php carbon's weekdays.
https://carbon.nesbot.com/docs/

This is useful for adding or subtracting business days,

Issue #307, opened 2018-11-11
struct.error: unpack requires a string argument of length 24


odinho:
Installed on an outdated `wheezy`. It did not have a proper `pip` so did a local install. Most other Python packages seems fine with this, but pendulum has a very weird error when trying to run:

```
$ python
Python 2.7.3 (default, Mar 13 2014, 11:03:55) 
[GCC 4.7.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pendulum
>>> pendulum.today()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pendulum/__init__.py", line 222, in today
    return now(tz).start_of('day')
  File "/usr/local/lib/python2.7/dist-packages/pendulum/__init__.py", line 207, in now
    dt = _datetime.datetime.now(local_timezone())
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/__init__.py", line 59, in local_timezone
    return get_local_timezone()
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/local_timezone.py", line 30, in get_local_timezone
    tz = _get_system_timezone()
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/local_timezone.py", line 58, in _get_system_timezone
    return _get_unix_timezone()
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/local_timezone.py", line 173, in _get_unix_timezone
    return Timezone(etctz.replace(' ', '_'))
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/timezone.py", line 30, in __init__
    tz = read(name, extend=extended)
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/zoneinfo/__init__.py", line 9, in read
    return Reader(extend=extend).read_for(name)
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/zoneinfo/reader.py", line 52, in read_for
    return self.read(file_path)
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/zoneinfo/reader.py", line 64, in read
    return self._parse(fd)
  File "/usr/local/lib/python2.7/dist-packages/pendulum/tz/zoneinfo/reader.py", line 90, in _parse
    hdr = self._parse_header(fd)
[...]

PR #411, opened 2019-10-11
Fix unpack error in Python 2.7.3

xq5he:
Ref https://github.com/sdispater/pendulum/issues/307
1. unpack doesn't accept a bytearray argument until Python 2.7.4 [issue10212](https://bugs.python.org/issue10212)
2. In python 2 bytes and str are same thing , there is not need to convert str to bytearray.


sdispater:
Python 2.7.3 is an old version of Python 2.7 so there is not need to support it.

Issue #333, opened 2019-01-16
TypeError: format() got an unexpected keyword argument 'formatter'


bolkedebruin:
```
>>> dt = pendulum.parse("2017-09-01 00:00:00+00:00")
>>> dt
DateTime(2017, 9, 1, 0, 0, 0, tzinfo=Timezone('+00:00'))
>>> dt.format('%Y-%m-%dT%H:%M:%S.%fZ', formatter='classic')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: format() got an unexpected keyword argument 'formatter'

```

Per https://github.com/sdispater/pendulum/issues/255 it is mentioned the keyword is not supported anymore, it then should be removed from the documentation.

sleao:
This should be fixed with a quick edit to [string_formatting.md](https://github.com/sdispater/pendulum/blob/master/docs/docs/string_formatting.md), am I right? If so, can I do it? If not, what else should I change in order to fix this?

PR #413, opened 2019-10-13
Removed formatter option from docs

LilJohny:
I removed option formatter from string_fromatting.md, which should fix issue #333


sdispater:
Already fixed on `master`.

PR #414, opened 2019-10-16
Default to own `.fold` when calling `.replace()`

tomage:
After doing a lot of experiments, it seemed to me that there was a problem when calling `.replace()` during DST off-transition (i.e. in the immediate hour after the transition). Reported issue here: https://github.com/sdispater/pendulum/issues/415.

So a seemingly innocent code like this:
```
    my_new_dt = my_old_dt.replace(microseconds=0)
```
would change the datetime in more ways than simply setting microseconds to zero.

This patch makes it so that when `.replace()` is called, it respects the `.fold` flag set on the datetime itself, which seems to resolve the issue.

Now, I'm no expert in timezones myself, so this might definitely be the exact wrong approach - but my gut feeling told me that it seemed odd to just default to the `pendulum.POST_TRANSITION` rule anytime we do replace, so I figured utilizing the flag seemed the right way to go about it. Seemed more symmetrical to the surrounding code too.

I added one more test module, to explicitly test out the `.replace()` function. Wasn't sure where best to fit it, but can easily accommodate a different place for it. As it happened, one of the "fluent" tests also broke - one that is dong a `.replace(tzinfo=...)` call. So I suppose I could have put my replace tests there, but wasn't sure what "fluent" meant in this context. So I ended up adding a few more cases for those `.replace(tzinfo=` tests, to get solid coverage surrounding DST transition points. Not quite convinced if those tests are what you had in mind when you wrote the original test, so hoping I'm not deviating from the intended design.

Well - looking forwards to your feedback - feel free to push back on any/all of this.

### DIFF
--- a/tests/datetime/test_fluent_setters.py
+++ b/tests/datetime/test_fluent_setters.py
@@ -139,8 +139,8 @@ def test_fluid_at_with_transition():
     assert 0 == new.second
 
 
-def test_replace_tzinfo_dls_off():
-    d = pendulum.datetime(2016, 3, 27, 0, 30)  # 30 min before DLS turning on
+def test_replace_tzinfo_dst_off():
+    d = pendulum.datetime(2016, 3, 27, 0, 30)  # 30 min before DST turning on
     new = d.replace(tzinfo=pendulum.timezone("Europe/Paris"))
 
     assert_datetime(new, 2016, 3, 27, 0, 30)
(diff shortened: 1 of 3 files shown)


tests/datetime/test_fluent_setters.py
```diff
@@ -139,18 +139,41 @@ def test_fluid_at_with_transition():
     assert 0 == new.second
 
 
-def test_replace_tzinfo():
-    d = pendulum.datetime(2016, 7, 2, 0, 41, 20)
+def test_replace_tzinfo_dls_off():
```
sdispater, line 142: Why dls here an not dst?

tomage: Oh right - I think I started off calling it "dls" (daylight-savings), but later saw that "dst" was a more common way to name it.

I'll rename to "dst".

tomage: Fixed with a fixup commit (i.e. I can squash into one commit before merge if you like, unless if you want to just squash+merge)

sdispater:
Thanks!

PR #419, opened 2019-10-25
Fix subtracting negative timedelta around DST

tomage:
Noticed that in a very specific case, subtracting a negative timedelta yielded different results than adding the same positive timedelta.

I.e. `my_dt + timedelta(hours=10) ~= my_dt - timedelta(hours=-10)`.

Digging into the git log, I found that a similar fix was done for the positive case here: https://github.com/sdispater/pendulum/commit/1ec1ad0dbdd34272d72fb750f82773f41fa8c8d7.

It seemed logical/symmetrical to do similar with the negative case, although there may have been some good reasons for why `_add_timedelta()` is different from `_subtract_timedelta()` that I'm missing.

I also threw in one test case to test specifically subtracting a negative timedelta.

Looking forward to your feedback as always.

### DIFF
--- a/pendulum/datetime.py
+++ b/pendulum/datetime.py
@@ -780,9 +780,7 @@ class DateTime(datetime.datetime, Date):
                 microseconds=delta.microseconds,
             )
 
-        return self.subtract(
-            days=delta.days, seconds=delta.seconds, microseconds=delta.microseconds
-        )
+        return self.subtract(seconds=delta.total_seconds())
 
     # DIFFERENCES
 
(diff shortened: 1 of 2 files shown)


sdispater:
Thanks for your contribution!

Can you rebase your PR on `master`?

tomage:
Sure thing! Just rebased on top of latest master from you.

(Just did the same for my other 2 PRs: https://github.com/sdispater/pendulum/pull/414 and https://github.com/sdispater/pendulum/pull/416)

sdispater:
Thanks for your contribution!

PR #427, opened 2019-12-12
fix: Perioid comparision with other classes and time delta and deprecation warnings

### LINKED ISSUE #421: Cannot check == between Period and timedelta
Hello, I'm having trouble comparing the result of subtracting two DateTimes with a `timedelta`. If I run the following snippet:

```py
import pendulum
from pendulum.period import Period
from datetime import timedelta
d1 = pendulum.datetime(2019, 1, 2, tz='UTC')
d2 = pendulum.datetime(2019, 1, 1, tz='UTC')
period = d1 - d2
print(period == timedelta(days=1))
```

I get the following error, instead of the `True` I was expecting:

```
AttributeError: 'datetime.timedelta' object has no attribute 'start'
```

wakemaster39:
Resolves #421 along with comparison against None and other classes.


wakemaster39:
this resolves #440

sdispater:
Thanks!

Issue #77, opened 2016-11-14
Adding Interval to Pendulum produces incorrect results


ulope:
```
from pendulum import Pendulum, Interval

p = Pendulum.now()
p2 = p + Interval(minutes=15)
assert p == p2
```

The reason is that `Pendulum.add_timedelta()` assumes a default python `timedelta` object and can't cope with the non-normalizing behavior of `Interval`.

sdispater:
The internals of Interval have been changed and fixed in the `develop` branch.

```python
>>> import pendulum
>>> p = pendulum.now()
>>> p2 = p + pendulum.interval(minutes=15)
>>> p == p2
False
```

This fix will land in the next `0.7.0` release.

PR #431, opened 2020-01-02
Fix RecursionError when adding to a DateTime with a FixedTimezone

### LINKED ISSUE #422: RecursionError when add or subtract-ing with a fixed offset timezone
The following:
```
import pendulum
pendulum.datetime(2019, 11, 1, tz=pendulum.tz.fixed_timezone(0)).subtract(seconds=1)
```
```
import dateutil, pendulum
pendulum.datetime(2019, 11, 1, tz=dateutil.tz.tzoffset(name=None, offset=0)).subtract(seconds=1)
```
```
import aniso8601, pendulum
pendulum.datetime(2019, 11, 1, tz=aniso8601.utcoffset.UTCOffset(minutes=0)).subtract(seconds=1)
```
all result in this RecursionError:
```
  ....
  File "/usr/local/lib/python3.8/site-packages/pendulum/tz/timezone.py", line 66, in convert
    return self._convert(dt)
  File "/usr/local/lib/python3.8/site-packages/pendulum/tz/timezone.py", line 334, in _convert
    return dt.astimezone(self)
  File "/usr/local/lib/python3.8/site-packages/pendulum/datetime.py", line 1466, in astimezone
    return pendulum.instance(super(DateTime, self).astimezone(tz))
  File "/usr/local/lib/python3.8/site-packages/pendulum/tz/timezone.py", line 345, in fromutc
    return (dt + self._utcoffset).replace(tzinfo=self)
  File "/usr/local/lib/python3.8/site-packages/pendulum/datetime.py", line 1442, in __add__
    return self._add_timedelta(other)
  File "/usr/local/lib/python3.8/site-packages/pendulum/datetime.py", line 750, in _add_timedelta
    return self.add(seconds=delta.total_seconds())
  File "/usr/local/lib/python3.8/site-packages/pendulum/datetime.py", line 660, in add
    dt = self.tz.convert(dt)
  File "/usr/local/lib/python3.8/site-packages/pendulum/tz/timezone.py", line 66, in convert
    return self._convert(dt)
  File "/usr/local/lib/python3.8/site-packages/pendulum/tz/timezone.py", line 334, in _convert
    return dt.astimezone(self)
  File "/usr/local/lib/python3.8/site-packages/pendulum/datetime.py", line 1466, in astimezone
    return pendulum.instance(super(DateTime, self).astimezone(tz))
  File "/usr/local/lib/python3.8/site-packages/pendulum/tz/timezone.py", line 345, in fromutc
    return (dt + self._utcoffset).replace(tzinfo=self)
  File "/usr/local/lib/python3.8/site-packages/pendulum/datetime.py", line 1442, in __add__
    return self._add_timedelta(other)
  File "/usr/local/lib/python3.8/site-packages/pendulum/datetime.py", line 750, in _add_timedelta
    return self.add(seconds=delta.total_seconds())
  File "/usr/local/lib/python3.8/site-packages/pendulum/datetime.py", line 621, in add
    offset = self.utcoffset()
RecursionError: maximum recursion depth exceeded
```

The actual date, offset, or number of seconds to add/subtract doesn't seem to matter, and using 'add' instead of subtract also causes the issue. A 'normal' timezone like 'Europe/London' won't cause the issue.

This came up with dates returned from our graphql API (which are parsed by graphene using aniso8601). Since all the fixed-offset tzinfo objects above trigger the same issue, I'm not sure what a good workaround would be, other than finding a 'normal' timezone that happens to have the same offset.

This is on CPython 3.8.0 on Linux, and I've tried with pendulum 2.0.4 and 2.0.5. I can't repro the issue on CPython 3.7.4, so I'm assuming it's 3.8-specific.

The closest existing issue I could find is https://github.com/sdispater/pendulum/issues/361 , but doesn't seem related - it's PyPy-specific, and `in_timezone` works okay on my example datetimes above.

pR0Ps:
This seems to have been caused by `bpo-32417`. Before that change, adding a timedelta to a date/datetime subclass would always return an instance of date/datetime instead of the subclass. After the change, the subclass is preserved.

The RecursionError was caused by adding a timedelta to a DateTime. Doing this uses the `convert` method of the DateTime's timezone to convert the new DateTime into the correct timezone. In the case of FixedTimezones, this requires adding the UTC offset of the timezone (a timedelta) to the DateTime, causing the recursion.

Before bpo-32417, the subclass of the DateTime was dropped while calling `astimezone`. This meant that the object that was passed into `fromutc` by `astimezone` was a stdlib datetime, not a Pendulum DateTime. Calling the stdlib datetime's add function would then do the addition and return the result (which would then be upconverted back into a Pendulum DateTime instance). Now, due to the subclass being preserved, the Pendulum DateTime's add function is being called instead, causing the recursion.

This commit fixes the RecursionError by always using the stdlib datetime's addition function to add the offset to the DateTime when calling `fromutc`.

bpo-32417: https://bugs.python.org/issue32417
commit: https://github.com/python/cpython/commit/89427cd0feae25bbc8693abdccfa6a8c81a2689c

Fixes #422

### DIFF
--- a/pendulum/tz/timezone.py
+++ b/pendulum/tz/timezone.py
@@ -342,7 +342,8 @@ class FixedTimezone(Timezone):
         return timedelta()
 
     def fromutc(self, dt):  # type: (datetime) -> datetime
-        return (dt + self._utcoffset).replace(tzinfo=self)
+        # Use the stdlib datetime's add method to avoid infinite recursion
+        return (datetime.__add__(dt, self._utcoffset)).replace(tzinfo=self)
 
     def tzname(self, dt):  # type: Optional[datetime]  # type: (...) -> Union[str, None]
         return self._name
--- a/tests/datetime/test_add.py
+++ b/tests/datetime/test_add.py
@@ -134,8 +134,9 @@ def test_addition_invalid_type():
 def test_add_to_fixed_timezones():
     dt = pendulum.parse("2015-03-08T01:00:00-06:00")
     dt = dt.add(weeks=1)
+    dt = dt.add(hours=1)
 
-    assert_datetime(dt, 2015, 3, 15, 1, 0, 0)
+    assert_datetime(dt, 2015, 3, 15, 2, 0, 0)
     assert dt.timezone_name == "-06:00"
     assert dt.offset == -6 * 3600
 


sdispater:
Thanks a lot for your contribution!

ErikBjare:
Will there be a new release that includes this PR anytime soon?

y0devv:
Duplicate of [PR #77](https://github.com/singer-io/tap-pipedrive/pull/77)

PR #439, opened 2020-01-21
Fix bug: week_of_month

dindom999:
There is a bug:

    In [1]: import pendulum

    In [2]: pendulum.parse("2020-01-01").week_of_month
    Out[2]: 1

    In [3]: pendulum.parse("2020-01-07").week_of_month
    Out[3]: 1  ==>should be 2

    In [4]: pendulum.parse("2020-01-14").week_of_month
    Out[4]: 2 ==>should be 3

### DIFF
--- a/pendulum/date.py
+++ b/pendulum/date.py
@@ -76,7 +76,13 @@ class Date(FormattableMixing, date):
 
     @property
     def week_of_month(self):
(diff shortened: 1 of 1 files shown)


StephenBrown2:
Looks like this was fixed in #446

PR #443, opened 2020-02-18
instantiation: correct the type of pendulum.instance

iomintz:
`pendulum.instance` receives a `datetime.datetime` instance, not a `pendulum.DateTime`.


sdispater:
Thanks!

PR #445, opened 2020-03-05
Fix pendulum.test to properly unwind after an exception

njsmith:
Previously, if an exception occurred inside a 'with pendulum.test(...)'
block, then the monkeypatch would remain in place instead of being
un-done. This could cause confusing results, as one failing test could
cause other tests to run with an unexpected mock in place and cause
other failures.


njsmith:
All the test failures appear to be unrelated.

sdispater:
Thanks!

PR #463, opened 2020-04-17
Add RFC7231 (HTTP Date) format

StephenBrown2:
This adds a date format string for RFC7231 (HTTP), and a convenience method to convert the datetime to `GMT` and output in the given format.

## Pull Request Check List

`develop` is severely out of date, so this is based off of `master`.

Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date
Examples:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After


Secrus:
Outdated

PR #464, opened 2020-04-17
Add timezone_abbr property shortcut to tzname()

StephenBrown2:
This adds a shortcut property to DateTime objects to get the timezone abbreviation, which can only be gotten from a datetime, since DST will change the result.

I had also wanted to get a list of acceptable abbreviations to add reverse parsing of "zz" in #463, but found it was more complicated than that and even getting the std abbreviation from a `Timezone` object directly wasn't possible.

I've added tests for everywhere `dt.timezone_name` is called, since it's a similar tied property, but if that's too many, I can back them out and just leave the one test in `tests/datetime/test_getters.py`

## Pull Request Check List


Secrus:
Outdated

Issue #467, opened 2020-05-01
Pendulum reports FileNotFound error when local timezone set with env var on Ubuntu


egbutter:
- [x] I am on the [latest](https://github.com/sdispater/pendulum/releases/latest) Pendulum version.
- [x] I have searched the [issues](https://github.com/sdispater/pendulum/issues) of this repo and believe that this is not a duplicate.

- Ubuntu 19.10 (eoan)
- Pendulum 2.1.0
- Timezone set via environment variable, `TZ=UTC`

## Issue
- Ran a test that relies on `pendulum`
- Test calls `pendulum.now().timestamp()`
- Expected: current seconds since epoch for local computer timezone
- Actual: traceback below - suggests that pendulum is trying to open a file with the name "UTC" instead of using the environment variable value "UTC" to set the timezone in `pendulum`

```
Traceback (most recent call last):
  File "/home/egbutter/.pyenv/versions/xxx/bin/pytest", line 10, in <module>
    sys.exit(main())
  File "/home/egbutter/.pyenv/versions/3.7.5/envs/xxx/lib/python3.7/site-packages/_pytest/config/__init__.py", line 125, in main
    config=config
  File "/home/egbutter/.pyenv/versions/3.7.5/envs/xxx/lib/python3.7/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/egbutter/.pyenv/versions/3.7.5/envs/xxx/lib/python3.7/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/egbutter/.pyenv/versions/3.7.5/envs/xxx/lib/python3.7/site-packages/pluggy/manager.py", line 87, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/home/egbutter/.pyenv/versions/3.7.5/envs/xxx/lib/python3.7/site-packages/pluggy/
  File "/home/egbutter/.pyenv/versions/3.7.5/envs/xxx/lib/python3.7/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/egbutter/.pyenv/versions/3.7.5/envs/xxx/lib/python3.7/site-packages/pluggy/callers.py", line 80, in get_result
    raise ex[1].with_traceback(ex[2])
[...]

PR #468, opened 2020-05-01
Adding try/except workaround when Pendulum tries to treat environment variable local timezone as a filepath

egbutter:
## Pull Request Check List

Addresses bug https://github.com/sdispater/pendulum/issues/467. Adding try/except workaround in `local_timezone.py` and test for the environment variable functionality. When `pendulum` incorrectly tries to treat an environment variable local timezone as a filepath instead of a string, carry on and try to treat the string as a timezone itself, instead.

- [x] Added **tests** for changed code - one new test illustrating env var behavior, asserting it still works
- [x] Updated **documentation** for changed code - added a single-line comment explaining change, nothing to add to docs
- [x] Based on `master` because this is a bugfix for current production `pendulum==2.1.0`

### DIFF
--- a/pendulum/tz/local_timezone.py
+++ b/pendulum/tz/local_timezone.py
@@ -248,7 +248,11 @@ def _tz_from_env(tzenv):  # type: (str) -> Timezone
 
     # TZ specifies a file
     if os.path.exists(tzenv):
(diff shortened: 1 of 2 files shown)


Secrus:
Outdated, superseded

PR #476, opened 2020-05-26
Fix markup at testing.md

Jamim:
Hello @sdispater,

> `Related methods will also return values mocked according to the **now** instance.`

Since this line is not a valid comment and it contains `**`, I believe it supposed to be outside of a code block.

Best regards!


sdispater:
Thanks!

PR #482, opened 2020-07-04
Fix the computation of total seconds with years and months

### LINKED ISSUE #475: Duration methods (total_days(), in_days()) return incorrect values and/or do not match docs

- **OS version and name**:  MacOS 10.14 (running Python 3.7.2)
- **Pendulum version**: 2.1.0

## Issue
The duration methods duration.total_days() and duration.in_days() do not match expected results, and/or do not match the docs.  Since the docs seem to show an inconsistent / incorrect result (specifically for duration.in_days()), it isn't clear what the real expected behavior is. 

I was quite confused about the logic and differences between the conversion of years/months into days for a duration (still am), so I just re-created the example used in the docs in order to test my thinking.  And when doing that, I got different results to what is shown in docs.

My results below, using exact example from Pendulum docs [duration section](https://pendulum.eustace.io/docs/#properties-and-duration-methods):

```python
# I copied example from pendulum docs
>>> it = pendulum.duration(
...     years=2, months=3,
...     days=1177, seconds=7284, microseconds=1234
... ) 

>>> it.days
1997   
# seems OK... 2 * 365 days/year + 3 * 30days/month + 1177 days = 1997

>>> it.total_days()
1177.0843055698379   
# this matches result shown in docs (1177.084...)
# but result seems counter-intuitive based on name of method
# I would expect 'total_days' to be 1997.084... including days from years and months

>>> it.in_days()
1177  
# this DOES NOT match result shown in docs (docs show 1997 for it.in_days())
# but does match description of this method being truncated integer of total_days()
```

Based on description of duration.in_days() method in the docs, I would expect duration.total_days() and duration.in_days() to return same result, only with in_days() as a truncated integer as described.  That's what I actually got, but doesn't match docs.

But larger issue seems to be that both .total_days() and .in_days() return incorrect (or at least un-intuitive) results -- I would expect them to both return 1997... in this case, including the days from years and months.  (not 1177)

Seems like there is either:
A) Incorrect example in the docs  (specifically where it shows return of 1997 for .in_days() method)
B) Bug in duration.in_days() and/or duration.total_days() method, especially when including years/months in duration
C) Both A and B
D)or I'm totally missing something...

This may be related to #387 but I'm creating this as separate bug because it seems there is at least a docs issue here, and because 387 relates to total_seconds() and maybe there are drop-in compatibility constraints with built-in total_seconds() method for that one?

Thanks to the creator & contributors!

### LINKED ISSUE #479: duration months cannot be added to other durations

- **OS version and name**: 
- **Pendulum version**: 

## Issue

    >>> pendulum.duration(weeks=1) + pendulum.duration(months=1) + pendulum.duration(days=1)
    Duration(weeks=1, days=1)

Note that adding months=1 has no effect.

sdispater:
Resolves #479 
Resolves #475 

## Pull Request Check List

### DIFF
--- a/pendulum/datetime.py
+++ b/pendulum/datetime.py
@@ -754,7 +754,7 @@ class DateTime(datetime.datetime, Date):
             )
         elif isinstance(delta, pendulum.Duration):
             return self.add(
-                years=delta.years, months=delta.months, seconds=delta.total_seconds()
+                years=delta.years, months=delta.months, seconds=delta._total
             )
 
         return self.add(seconds=delta.total_seconds())
@@ -770,14 +770,7 @@ class DateTime(datetime.datetime, Date):
         """
         if isinstance(delta, pendulum.Duration):
             return self.subtract(
-                years=delta.years,
-                months=delta.months,
-                weeks=delta.weeks,
-                days=delta.remaining_days,
-                hours=delta.hours,
-                minutes=delta.minutes,
-                seconds=delta.remaining_seconds,
-                microseconds=delta.microseconds,
+                years=delta.years, months=delta.months, seconds=delta._total
             )
 
         return self.subtract(seconds=delta.total_seconds())
--- a/pendulum/duration.py
+++ b/pendulum/duration.py
@@ -66,11 +66,19 @@ class Duration(timedelta):
             raise ValueError("Float year and months are not supported")
 
         self = timedelta.__new__(
-            cls, days, seconds, microseconds, milliseconds, minutes, hours, weeks
+            cls,
+            days + years * 365 + months * 30,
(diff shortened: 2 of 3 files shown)


jacobg:
This commit breaks the following:
```python
>>> dt = pendulum.DateTime(2022, 6, 8, 13, 21, 12, 239829, tzinfo=pendulum._Timezone('America/New_York'))
>>> dt._add_timedelta(datetime.timedelta(seconds=3600))
DateTime(2022, 6, 8, 14, 21, 12, 239829, tzinfo=Timezone('America/New_York'))
```

The _add_timedelta call is a no-op.

PR #485, opened 2020-07-16
 Expose Timezone._utcoffset property for pandas compatibility

### LINKED ISSUE #131: Pandas compatibility of Timezone / Timezone Info
I think there is an incompatibility between ```pandas``` and ```pendulum.tz.timezone.Timezone```. I'm not sure who is to blame though, but as from ```pendulum.tz.timezone_info.TimezoneInfo``` it seemed reasonable to check here first.

Two examples of unexpected behaviour; the programs exit with a segmentation fault:

```python
from datetime import datetime, timedelta

from pendulum.tz.timezone import UTCTimezone
from pendulum.tz.timezone_info import UTC
import pandas as pd
import pendulum
```

This program creates two date time ranges, but printing (creating a string representation) fails for the second:
```python
tomorrow = pendulum.tomorrow()

print(pd.date_range(today.astimezone('UTC'), tomorrow.astimezone('UTC'), freq='1H'))
print(pd.date_range(today.astimezone('CET'), tomorrow.astimezone('CET'), freq='1H'))  # seg faults
```

In this example using UTCTimezone causes the segmentation fault:
```python
tomorrow = datetime.now() + timedelta(hours=24)
now = datetime.now()

r = pd.date_range(now, tomorrow, freq='1H')
r.tz = UTC
print(r[0])
r.tz = UTCTimezone
print(r[0]) # seg faults
```

I recon things go wrong when boxing in pandas. This is in C / Cython territory, so lot harder to debug for me.

Anyone ideas on the matter? I'd say ```Timezone``` probably behaves a bit different from ```TimezoneInfo``` but I really don't know where to start. I haven't even found their respective use cases yet.

Versions used:
 - pendulum==1.2.4
 - pandas==0.20.2
 - numpy==1.13.0

I'm running on Fedora Linux.

Thanks in advance!

ddelange:
This commit allows pandas to get _utcoffset from a Timezone object, taking the transision as seen from utcnow. 

As explained by @sdispater in the issues linked below, pandas tries to access the `Timezone._utcoffset` property (which isn't currently implemented for variable Timezone objects) and subsequently tries to get `Timezone.utcoffset(None)` (which correctly returns None as it can't be determined without knowlede of date and time). This PR allows pandas to treat Timezones as if they were FixedTimezones (as returned for utcnow).

ref:
https://github.com/pandas-dev/pandas/blob/v1.0.5/pandas/_libs/tslibs/timezones.pyx#L152

Fixes https://github.com/pandas-dev/pandas/issues/15986
Closes #131 

```py
>>> import pendulum
>>> import pandas
>>> pandas.date_range(start=pendulum.yesterday(), end=pendulum.tomorrow())
DatetimeIndex(['2020-07-15 00:00:00+02:00', '2020-07-16 00:00:00+02:00',
               '2020-07-17 00:00:00+02:00'],
              dtype='datetime64[ns, Timezone('Europe/Berlin')]', freq='D')
```

## Pull Request Check List

> - [ ] Updated **documentation** for changed code.

Left this unticked, as it's a private property that users won't touch.

### DIFF
--- a/pendulum/tz/timezone.py
+++ b/pendulum/tz/timezone.py
@@ -234,6 +234,10 @@ class Timezone(tzinfo):
 
         return transition.utcoffset()
 
+    @property
+    def _utcoffset(self):  # type: () -> timedelta
+        return self.utcoffset(datetime.utcnow())
+
     def dst(
         self, dt  # type: Optional[_datetime]
     ):  # type: (...) -> Optional[timedelta]
@@ -309,11 +313,14 @@ class FixedTimezone(Timezone):
 
         self._name = name
         self._offset = offset
-        self._utcoffset = timedelta(seconds=offset)
 
     @property
     def offset(self):  # type: () -> int
         return self._offset
+
+    @property
+    def _utcoffset(self):  # type: () -> timedelta
+        return timedelta(seconds=self.offset)
 
     def _normalize(self, dt, dst_rule=None):  # type: (_D, Optional[str]) -> _D
         if _HAS_FOLD:
--- a/tests/tz/test_timezone.py
+++ b/tests/tz/test_timezone.py
@@ -585,6 +585,12 @@ def test_timezones_extension_can_be_disabled():
     assert dt.dst() == timedelta()
 
 
+def test_timezone_has_private_utcoffset_attribute():
+    now = pendulum.now()
+    pandas_call = now.tzinfo._utcoffset.total_seconds()
+
+    assert now.utcoffset().total_seconds() == pandas_call
+
 def test_repr():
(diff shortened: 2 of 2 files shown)


ddelange:
Closing this as it's quite incorrect behaviour

philsheard:
It doesn't appear that this has been addressed on either library, so if anyone wants a workaround then this worked for me:

Do a roundtrip from Pendulum > ISO format > Pandas to avoid the internal lookup that causes the warning.

`converted_date = pandas.to_datetime(pendulum.today().isoformat())`

Not pretty but it will clean your error logs without hurting comprehension.

Issue #470, opened 2020-05-03
Unable to install pendulum with --no-binary :all: (proxy for alpine or such where there aren't wheels)


altendky:
- **OS version and name**: Ubuntu 19.10
- **Pendulum version**: 2.1.0

## Issue

This is a new error while I was looking back at #458, #454, #457, etc.

```bash
altendky@p1:~/repos/pendulum/tmp$ python3.8 -m venv venv
```
```bash
altendky@p1:~/repos/pendulum/tmp$ venv/bin/pip install --upgrade pip setuptools
Collecting pip
  Using cached https://files.pythonhosted.org/packages/54/2e/df11ea7e23e7e761d484ed3740285a34e38548cf2bad2bed3dd5768ec8b9/pip-20.1-py2.py3-none-any.whl
Collecting setuptools
  Using cached https://files.pythonhosted.org/packages/a0/df/635cdb901ee4a8a42ec68e480c49f85f4c59e8816effbf57d9e6ee8b3588/setuptools-46.1.3-py3-none-any.whl
Installing collected packages: pip, setuptools
  Found existing installation: pip 19.2.3
    Uninstalling pip-19.2.3:
      Successfully uninstalled pip-19.2.3
  Found existing installation: setuptools 41.2.0
    Uninstalling setuptools-41.2.0:
      Successfully uninstalled setuptools-41.2.0
Successfully installed pip-20.1 setuptools-46.1.3
```
```bash
altendky@p1:~/repos/pendulum/tmp$ venv/bin/pip install --no-binary :all: pendulum
Collecting pendulum
  Using cached pendulum-2.1.0.tar.gz (80 kB)
  Installing build dependencies ... error
  ERROR: Command errored out with exit status 1:
   command: /home/altendky/repos/pendulum/tmp/venv/bin/python3.8 /home/altendky/repos/pendulum/tmp/venv/lib/python3.8/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-z78_ptnj/overlay --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- 'poetry>=1.0.0b1'
       cwd: None
  Complete output (139 lines):
  Collecting poetry>=1.0.0b1
    Using cached poetry-1.1.0a1.tar.gz (107 kB)
    Installing build dependencies: started
    Installing build dependencies: finished with status 'done'
    Getting requirements to build wheel: started
    Getting requirements to build wheel: finished with status 'done'
[...]

PR #487, opened 2020-07-22
Add setuptools to the build-system requires

hesstobi:
With version 2.1.1 the installing with `pip install --no-binary pendulum pendulum` is failing, with the error `ModuleNotFoundError: No module named 'setuptools'`. This was highligtet in #470. I was able to fix this by adding `setuptools` to the `[build-system]` `requires`.

## Pull Request Check List

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,5 +69,5 @@ known_third_party = [
 
 
 [build-system]
-requires = ["poetry>=1.0.0b1"]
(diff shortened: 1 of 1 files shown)


benkonrath:
This is working for me. 👍 Thanks!

benkonrath:
It seems that this has been fixed in master and the 2.1 branch so this is now unnecessary.

PR #511, opened 2020-11-01
tr locale added

hakkikonu:
I added tr locale and its tests.


hakkikonu:
On Linting step I get **Error: Process completed with exit code 1.** But where is the problem? @sdispater

PR #546, opened 2021-05-26
Add type hints for formatting

traviscook21:
When using mypy with pendulum, I run into issues like:
```
Call to untyped function "format" in typed context
```

Mypy doesn't process types in docstrings and instead preferences comments like https://github.com/sdispater/pendulum/pull/320/files

This should fix the issue for calls to `.format`

## Pull Request Check List


Secrus:
This was done in the new version of typing. Closing as no longer relevant.

PR #551, opened 2021-06-13
mypy: implicit_reexport

altendky:
Either this or an explicit `__all__` is needed to be able to use type hints in a strict setting without the using application setting this up themselves in their own `mypy.ini`.  As is there are errors such as `error: Name "pendulum.DateTime" is not defined  [name-defined]`.


Secrus:
The problem was resolved in a different PR.

Issue #608, opened 2022-03-09
`with pendulum.test(...)` gets mangled by timezone issues


rileyjohngibbs:
- [x] I am on the [latest](https://github.com/sdispater/pendulum/releases/latest) Pendulum version.
- [x] I have searched the [issues](https://github.com/sdispater/pendulum/issues) of this repo and believe that this is not a duplicate.

- **OS version and name**: macOS 11.6.1
- **Pendulum version**: 3.0.0a0

## Issue

When using the `pendulum.test` context in tests, the local timezone is getting set to Toronto when using `pendulum.today()`. To see this behavior, add this test to `tests/date/test_diff.py`:

```python
def test_pendulum_test():
    with pendulum.test(pendulum.datetime(2016, 3, 1)):
        today = pendulum.today().date()
        assert today == pendulum.datetime(2016, 3, 1).date()
```

Alternatively, try:

```python
def test_pendulum_test():
    with pendulum.test(pendulum.datetime(2016, 3, 1)):
        assert str(pendulum.now().tz) == "UTC"
        assert str(pendulum.today().tz) == "UTC"
```

You can see that `.now()` returns a UTC timezone but `.today()` returns Toronto. (Or at least, this is what I'm seeing, even though I'm not in the Toronto timezone, I'm in the Los Angeles timezone.) As a result, `pendulum.today().date()` evaluates to (timezone-naive) February 29, because it's still February in Toronto when it's midnight on March 1 in UTC.

You can see this in a REPL too:
```
>>> import pendulum
>>> with pendulum.test(pendulum.datetime(2016, 3, 1)):
...     print(pendulum.today().date())
... 
2016-02-29
```

This is occurring in some existing tests in `test_diff`, for example `test_diff_for_humans_now_and_month`, but seems it is being masked by another bug (see #606) that causes the time difference to be calculated incorrectly (or at the very least inconsistently) for months that do not have 31 days.

alejandro-angulo:
I think the Toronto timezone you're noticing is related to the setup fixture https://github.com/sdispater/pendulum/blob/master/tests/conftest.py#L8

Though, I guess it is a bit strange that the timezone isn't consistent (`now` uses UTC but `today` uses what's set by `set_local_timezone`).

Issue #606, opened 2022-03-09
in_months behavior inconsistent in 30-day months


rileyjohngibbs:
- [x] I am on the [latest](https://github.com/sdispater/pendulum/releases/latest) Pendulum version.
- [x] I have searched the [issues](https://github.com/sdispater/pendulum/issues) of this repo and believe that this is not a duplicate.

- **OS version and name**: macOS 11.6.1
- **Pendulum version**: 2.1.2 and 3.0.0a0

## Issue

When calculating a diff's `in_months` method, the following results seem inconsistent:

| start | end | `in_months` result |
| --- | --- | --- |
| `2022-01-02` | `2023-01-01` | `11` |
| `2022-02-02` | `2023-02-01` | `11` |
| `2022-03-02` | `2023-03-01` | `11` |
| `2022-04-02` | `2023-04-01` | `12` |
| `2022-05-02` | `2023-05-01` | `11` |
| `2022-06-02` | `2023-06-01` | `12` |
| `2022-07-02` | `2023-07-01` | `11` |
| `2022-08-02` | `2023-08-01` | `11` |
| `2022-09-02` | `2023-09-01` | `12` |
| `2022-10-02` | `2023-10-01` | `11` |
| `2022-11-02` | `2023-11-01` | `12` |
| `2022-12-02` | `2023-12-01` | `11` |

It seems that whenever the end date is in one of the months with 30 days (April, June, September, November), if the start date's day of the month is one day later than the corresponding end day of the month, the number of months from `in_months` is one higher than it "should" be.

Obviously this is a sticky issue because the definition of a "month" is not well defined (it can be any value from 28 to 31 days), but since this behavior doesn't happen with February, the behavior seems inconsistent at the least.

### Desired Result

The `in_months` method should be determined based on the comparison of the day of the month, not the duration of the last month in the interval, which seems to be [the root](https://github.com/sdispater/pendulum/blob/f1df7dc3f838bd4ab1075ba25c8b6ce5d8141995/pendulum/_extensions/_helpers.c#L827) of this behavior.

rileyjohngibbs:
Some more inconsistency, specifically with February:
```
>>> import pendulum
>>> start = pendulum.parse("2022-02-05")
>>> for d in range(1, 29):
...     months = (pendulum.parse(f"2023-02-{str(d).zfill(2)}") - start).in_months()
...     print(d, months)
... 
1 11
2 12
3 11
4 11
5 12
6 12
7 12
8 12
9 12
[...]

PR #609, opened 2022-03-10
In months fix

rileyjohngibbs:
## Pull Request Check List

This is to address issue #606. The change to the `setup` fixture in the `test_diff` module to use UTC instead of Toronto is a workaround for issue #608.

Not sure whether this should be based on `develop` or `master`. Some of this is clearly a bug (see February in particular), but some of the side effects are changes in user-facing behavior of what pendulum thinks "1 month" means colloquially, e.g. whether March 31st to April 30th is 1 month (forward), and whether April 30th to March 31st is 1 month (backward).

Rather than try to tweak the logic to satisfy the current notions of what "1 month" means, I let them change and updated the tests to match the new behavior.

### DIFF
--- a/pendulum/_extensions/_helpers.c
+++ b/pendulum/_extensions/_helpers.c
@@ -824,14 +824,6 @@ PyObject *precise_diff(PyObject *self, PyObject *args)
                 day_diff += days_in_last_month;
             }
         }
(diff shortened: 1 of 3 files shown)


Secrus:
Outdated

PR #616, opened 2022-04-16
Fix from_format using local timezone even when timezone is provided.

### LINKED ISSUE #613: breaking: from_format errors out when there is no local timezone configuration but timezone is passed in

- **OS version and name**: This is a docker container, so the output of `cat /proc/version` is this: Linux version 5.4.0-1071-aws (buildd@lcy02-amd64-020) (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04))
- **Pendulum version**: 2.1.2

## Issue
In a docker container with no timezone configuration (presumably), `from_format` errors out when passing in the timezone parameter. For example, this line from the docs: `pendulum.from_format('1975-05-21 22', 'YYYY-MM-DD HH', tz='Europe/London')` will spit out the error `RuntimeError: Unable to find any timezone configuration`. This same behavior happens when I try to run something like `pendulum.now()` with no arguments which makes sense because I'm assuming it's trying to get the local time but can't. In our docker container we strictly do not want anything to use local time when it's not needed which is why there is no timezone configuration and I feel like it's completely unnecessary here. If I pass in the `tz` parameter it shouldn't need to rely on local time at all. Looking at the source code, I'm pretty sure this is because `now()` is being used here: https://github.com/sdispater/pendulum/blob/f1df7dc3f838bd4ab1075ba25c8b6ce5d8141995/pendulum/__init__.py#L238. I'd love to use pendulum in our production server given the ease of managing timezones, but this behavior is completely blocking it and rendering it unusable right now.

ayoung19:
## Pull Request Check List

Fixes #613

### DIFF
--- a/pendulum/__init__.py
+++ b/pendulum/__init__.py
@@ -235,7 +235,7 @@ def from_format(
     """
     Creates a DateTime instance from a specific format.
     """
-    parts = _formatter.parse(string, fmt, now(), locale=locale)
+    parts = _formatter.parse(string, fmt, now(tz), locale=locale)
     if parts["tz"] is None:
         parts["tz"] = tz
 


ayoung19:
NOTICE: Just a draft, need to work out the expected behavior with @sdispater as this new change defaults timezone to UTC when no timezone is provided (unlike locally before).

jamwine (approved):
The timezone within the `now()` will create a DateTime instance in accordance with the proper timezone.

ayoung19:
this is potentially breaking, but it is useful for how i'm using it at my company. do u think this is the correct behavior? it shouldnt be relying on the local timezone of the machine right?

PR #633, opened 2022-08-02
feat: Use importlib.resources to load package files

### LINKED ISSUE #632: Use `importlib.resources` to load `tzdata` files and replace use of `__file__`
- [x] I have searched the [documentation](https://pendulum.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

From the [`importlib-resources` docs](https://importlib-resources.readthedocs.io/en/latest/using.html):

> The use of `__file__` doesn’t work if your package lives inside a zip file, since in that case this code does not live on the file system.

This is the case also for applications that produce binaries. From [the PyOxidizer docs](https://pyoxidizer.readthedocs.io/en/stable/oxidized_importer_behavior_and_compliance.html#no-file):

> These attributes are not set because it isn’t obvious what the values should be! Typically, `__file__` is used by Python as an anchor point to derive the path to some other file. However, when loading modules from memory, the traditional filesystem hierarchy of Python modules does not exist. In the opinion of PyOxidizer’s maintainer, exposing `__file__` would be lying and this would cause more potential for harm than good.

I know this is a somewhat niche use case, but the fix is relatively simple and safe. Furthermore, `importlib.resources` is now fairly stable with [clear deprecations for Python 3.11](https://docs.python.org/3.11/library/importlib.resources.html#deprecated-functions).

I can submit a PR if there's interest.

edgarrmondragon:
## Pull Request Check List

Closes #632

### DIFF
--- a/pendulum/locales/locale.py
+++ b/pendulum/locales/locale.py
@@ -1,11 +1,16 @@
-import os
 import re
+import sys
 
 from importlib import import_module
 from typing import Any
 from typing import Dict
 from typing import Optional
 from typing import Union
+
+if sys.version_info >= (3, 9):
+    from importlib import resources
+else:
+    import importlib_resources as resources
 
 
 class Locale:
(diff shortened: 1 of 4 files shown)


Secrus:
Could you please sync changes from master branch? CI config was changed.

edgarrmondragon:
> Could you please sync changes from master branch? CI config was changed.

@Secrus done!

PR #670, opened 2022-11-09
Fix typing of Timezone and ZoneInfo

bryanforbes:
Currently the method used to import `zoneinfo` leaves it typed as `Any` in `mypy` and `Unknown` in `pyright`. This PR updates `pendulum.utils._compat` to inform type checkers of the correct type information for `zoneinfo` while leaving the import mechanism the same during runtime. With `Timezone` having the correct type info from `ZoneInfo`, a few more type errors surfaced which I took care of either by adding/removing `# type: ignore` or adding/removing `cast()` calls. Since no logic has changed at runtime, no unit tests were added.

### DIFF
--- a/pendulum/utils/_compat.py
+++ b/pendulum/utils/_compat.py
@@ -5,9 +5,9 @@ import sys
 PYPY = hasattr(sys, "pypy_version_info")
 PY38 = sys.version_info[:2] >= (3, 8)
 
-try:
+if sys.version_info < (3, 9):
     from backports import zoneinfo
-except ImportError:
-    import zoneinfo  # type: ignore[no-redef]
+else:
+    import zoneinfo
 
 __all__ = ["zoneinfo"]
--- a/pendulum/_extensions/helpers.py
+++ b/pendulum/_extensions/helpers.py
@@ -353,7 +353,7 @@ def _get_tzinfo_name(tzinfo: datetime.tzinfo | None) -> str | None:
 
     if hasattr(tzinfo, "key"):
         # zoneinfo timezone
-        return cast(str, cast(zoneinfo.ZoneInfo, tzinfo).key)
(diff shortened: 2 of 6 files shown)


pendulum/utils/_compat.py
```diff
@@ -2,12 +2,20 @@
...
+from typing import TYPE_CHECKING
+
 PYPY = hasattr(sys, "pypy_version_info")
 PY38 = sys.version_info[:2] >= (3, 8)
 
-try:
-    from backports import zoneinfo
-except ImportError:
-    import zoneinfo  # type: ignore[no-redef]
+if TYPE_CHECKING:
```
Secrus, line 10: will having only the `TYPE_CHECKING` part satisfy `mypy`? we are using `pyupgrade`, which understands this type of syntax (unlike try-ImportError) and handles it when upgrading.

bryanforbes: > will having only the `TYPE_CHECKING` part satisfy `mypy`? we are using `pyupgrade`, which understands this type of syntax (unlike try-ImportError) and handles it when upgrading.

If you are asking if the whole `if TYPE_CHECKING/else` block can be simplified to just the version check condition, the answer is yes. That whole block can be simplified to:

```python
if sys.version_info < (3, 9):
    from backports import zoneinfo
else:
    import zoneinfo
```

This will work at runtime and during type checking (in both `mypy` and `pyright` and I would suspect other type checkers as well). IMO the simplified version is the better solution, but I was trying to make as few changes to the runtime behavior as possible.

Secrus: Go with this solution, please. We are going into 3.0 anyway, so it's ok if there are some changes.

bryanforbes: I've updated PR to address this

Secrus (approved):

PR #676, opened 2022-11-20
Rename the Period class to Interval

sdispater:
## Pull Request Check List

This change is both semantic change and breaking change.

This renames the `Period` class to `Interval` to make its purpose and what it represents clearer.


pendulum/date.py
```diff
@@ -283,7 +283,7 @@ def __sub__(self, other: timedelta | date) -> Date | Period:
 
     # DIFFERENCES
 
-    def diff(self, dt: date | None = None, abs: bool = True) -> Period:
+    def diff(self, dt: date | None = None, abs: bool = True) -> Interval:
         """
         Returns the difference between two Date objects as a Period.
```
Secrus, line 288:
```suggestion
        Returns the difference between two Date objects as an Interval.
```

pendulum/__init__.py
```diff
@@ -286,11 +286,11 @@ def duration(
     )
 
 
-def period(start: DateTime, end: DateTime, absolute: bool = False) -> Period:
+def interval(start: DateTime, end: DateTime, absolute: bool = False) -> Interval:
     """
     Create a Period instance.
```
Secrus, line 291:
```suggestion
    Create an Interval instance.
```

Secrus (changes requested):

pendulum/date.py
sdispater: Done!

pendulum/__init__.py
sdispater: Done!

Secrus (approved):

PR #677, opened 2022-11-23
Mypy improvement

Secrus:
## Pull Request Check List

- [ ] Added **tests** for changed code. **Not needed**
- [ ] Updated **documentation** for changed code. **No changes require documenting**

Mypy 0.990 added `show_error_codes` as enabled by default and our `# type: ignore` comments had wrong codes which caused a failure in #674. 

Second commit moves `mypy` usage away from `pre-commit`, since when it's being run from `pre-commit`, it might miss some typing issues coming from dependencies. See https://github.com/python-poetry/cleo/pull/254#issuecomment-1267690650

### DIFF
--- a/pendulum/datetime.py
+++ b/pendulum/datetime.py
@@ -41,7 +41,7 @@ from pendulum.tz.timezone import Timezone
 from pendulum.utils._compat import PY38
 
 if TYPE_CHECKING:
-    from typing import Literal
+    from pendulum.utils._compat import Literal
 
 
 class DateTime(datetime.datetime, Date):
@@ -1349,7 +1349,7 @@ class DateTime(datetime.datetime, Date):
     ]:
         return self.__reduce_ex__(2)
 
-    def __reduce_ex__(  # type: ignore[override]
+    def __reduce_ex__(
         self, protocol: int
     ) -> tuple[
         type[DateTime], tuple[int, int, int, int, int, int, int, datetime.tzinfo | None]
--- a/pendulum/time.py
+++ b/pendulum/time.py
@@ -19,7 +19,7 @@ from pendulum.duration import Duration
 from pendulum.mixins.default import FormattableMixin
 
 if TYPE_CHECKING:
-    from typing import Literal
+    from pendulum.utils._compat import Literal
 
 
 class Time(FormattableMixin, time):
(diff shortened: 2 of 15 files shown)


pendulum/utils/_compat.py
```diff
@@ -6,8 +6,20 @@
...
     from backports import zoneinfo
 else:
     import zoneinfo
 
-__all__ = ["zoneinfo"]
+    from importlib import resources
+
+
+if sys.version_info < (3, 8):
+    from typing_extensions import Literal
```
bryanforbes, line 19: [Aliasing things like `Literal`](https://github.com/microsoft/pyright/issues/3025#issuecomment-1035266830) is not supported in pyright, so this one should probably be done inline where it is needed

pendulum/utils/_compat.py
```diff
@@ -6,8 +6,20 @@
...
+    from typing_extensions import SupportsIndex
```
bryanforbes, line 20: This makes `typing_extensions` a runtime dependency. Is that desired?

pendulum/datetime.py
```diff
@@ -41,7 +41,7 @@
 from pendulum.utils._compat import PY38
 
 if TYPE_CHECKING:
-    from typing import Literal
+    from pendulum.utils._compat import Literal
```
bryanforbes, line 44: This should be changed to `from typing_extensions import Literal` so that pyright is supported

pendulum/time.py
```diff
@@ -19,7 +19,7 @@
 from pendulum.mixins.default import FormattableMixin
 
 if TYPE_CHECKING:
-    from typing import Literal
+    from pendulum.utils._compat import Literal
```
bryanforbes, line 22: This should be changed to `from typing_extensions import Literal` so that pyright is supported

pendulum/utils/_compat.py
Secrus: We are only supporting `mypy` check in our pipelines right now and I stick to how `mypy` works (never worked with `pyright`, but it sounds like it's not compatible with `mypy`).

pendulum/utils/_compat.py
Secrus: Probably not :laughing:. I will need to rework this PR anyway since typing fails depending on the version of Python it's run on.

pendulum/utils/_compat.py
bryanforbes: `pyright` and `mypy` work pretty hard to be compatible with one another. `pyright` does an early scan for some special forms to speed things up, and the `Literal` import is one of those. This is one of the few differences (aside from `pyright` supporting more PEPs) between `pyright` and `mypy`. The other thing to take into account is that `pyright` is the type checker that is included by default with Visual Studio Code's Python extension.

PR #678, opened 2022-11-23
Release 3.0.0a1

sdispater:
### Added

- Added new testing helpers to time travel. [#626](https://github.com/sdispater/pendulum/pull/626)

### Changed

- Dropped support for Python 2.7, 3.5 and 3.6. [#569](https://github.com/sdispater/pendulum/pull/569)
- The `Timezone` class now relies on the native `zoneinfo.ZoneInfo` class. [#569](https://github.com/sdispater/pendulum/pull/569)
- Renamed the `Period` class to `Interval`. [#676](https://github.com/sdispater/pendulum/pull/676)
- Renamed the `period` helper to `interval`. [#676](https://github.com/sdispater/pendulum/pull/676)
- Removed existing testing helpers: `test()` and `set_test_now()`. [#626](https://github.com/sdispater/pendulum/pull/626)

### Locales

- Added the `sk` locale. [#575](https://github.com/sdispater/pendulum/pull/575)
- Added the `ja` locale. [#610](https://github.com/sdispater/pendulum/pull/610)
- Added the `he` locale. [#585](https://github.com/sdispater/pendulum/pull/585)
- Added the `sv` locale. [#562](https://github.com/sdispater/pendulum/pull/562)

Issue #270, opened 2018-08-28
rfc1123 issues


trancee:
I am using *pendulum==2.0.3* and I want to return an RFC-1123 compliant date in the HTTP response headers as defined here:
[HTTP-date](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.3.1)

The date is formatted like this (today's date):
`Mon, 27 Aug 2018 02:13:09 +0000`

Now I will feed that date back into my code and Pendulum should parse it, but instead throws an exception:
> pendulum.parsing.exceptions.ParserError: Unable to parse string [Mon, 27 Aug 2018 02:13:09 +0000]

According to the documentation which I shared the link above, the date should be formatted something like this:
> Sun, 06 Nov 1994 08:49:37 GMT  ; RFC 822, updated by RFC 1123

How come Pendulum produces a date format which it can't parse itself? 
And is there proper support for RFC 1123 and RFC 822 in Pendulum or do I need to use something else?

Thanks for your help and guidance.

sdispater:
You can take two approaches here. You can either use the `from_format()` helper:

```python
>>> import pendulum
>>> pendulum.from_format("Mon, 27 Aug 2018 02:13:09 +0000", "ddd, DD MMM YYYY HH:mm:ss ZZ")
DateTime(2018, 8, 27, 2, 13, 9, tzinfo=Timezone('+00:00'))
```

or use the `parse()` method in non-strict mode:

```python
>>> import pendulum
>>> pendulum.parse("Mon, 27 Aug 2018 02:13:09 +0000", strict=False)
DateTime(2018, 8, 27, 2, 13, 9, tzinfo=Timezone('+00:00'))
```

trancee:
Thank you for the tips. I chose the second approach and that seems to be working.

But for my first question, I am still not sure. The referred specification about HTTP-date says that the timezone should be *GMT* whereas Pendulum returns *+0000*.

> Mon, 27 Aug 2018 02:13:09 +0000

Is it possible that Pendulumn returns *GMT* in place of *+0000* to be compliant with the standards of *rfc1123-date* and *rfc850-date*?

sdispater:
From what I read in the RFC it says:

> There is a strong trend towards the use of numeric timezone
         indicators, and implementations SHOULD use numeric timezones
         instead of timezone names.  However, all implementations MUST
         accept either notation.  If timezone names are used, they MUST
         be exactly as defined in RFC-822.

[...]

Issue #222, opened 2018-05-30
timestamp with miliseconds


jgrammen-agilitypr:
I would like to print unix time stamps (time since epoch) with milliseconds precision, there does not appear to be a function for this in the library.
Is there anyway to get a date in milliseconds since epoch?

sdispater:
The `DateTime.float_timestamp` property gives you a timestamp with milliseconds precision (see https://pendulum.eustace.io/docs/#attributes-and-properties).

jgrammen-agilitypr:
that actually gives you the timestamp in microsecond precision, as of the posting of this comment the timestamp since epoch in miliseconds is 
```
1529085780669
```
which has 13 digits.

the pendulum function in the example returns
```
>>> dt.float_timestamp
1346887571.123789
```
which has 16 digits (ignoring the period)

So guess the solution is to multiple by 1000 which gets me miliseconds and then drop everything after the decimal (the microsecond part).

Issue #188, opened 2018-03-24
Feature request: periods with a 'human' interval


bolkedebruin:
Many interval / periods are determined by business requirements. So for example "run this at every first monday of the month at 9.00am UTC" or "last friday of the month  . At the moment this is impossible to do with pendulum (or with croniter for that matter).

So I would like to suggest a new feature that enabled this. I'm just not sure how to name it ;-).

PR #679, opened 2022-11-26
Switch `METH_VARGS` to `METH_O`

movermeyer:
### What are you trying to accomplish?

While working on `ciso8601`, I [discovered](https://github.com/closeio/ciso8601/pull/130) that things run a lot faster if you define your methods as [`METH_O`](https://docs.python.org/3/c-api/structures.html#METH_O) (i.e., method with a single object argument) you can skip the overhead of `PyArg_ParseTuple`.

### What approach did you choose and why?

Switched the instances of `METH_VARGS` to be `METH_O` instead.

### The impact of these changes

| Item         | Performance of `parse_iso8601` |
|--------------|-----------|
| `master`     | 1000000, 0.2697502020164393 => #270 nsec   |
| This PR      | 1000000, 0.22211680599139072 => #222 nsec  |
| v2.1.2       | 2000000, 0.3753614659945015 => #188 nsec   |

This PR shaves ~48 ns (17%) off the runtime of `parse_iso8601`.

### Testing

Here's the script I used.

```python
import timeit

UNITS = {"nsec": 1e-9, "usec": 1e-6, "msec": 1e-3, "sec": 1.0}
SCALES = sorted([(scale, unit) for unit, scale in UNITS.items()], reverse=True)

def format_duration(duration):
    # Based on cPython's `timeit` CLI formatting
    scale, unit = next(((scale, unit) for scale, unit in SCALES if duration >= scale), SCALES[-1])
    precision = 3
    return "%.*g %s" % (precision, duration / scale, unit)


timer = timeit.Timer(stmt="parse_iso8601('2014-01-09T21:48:00')", setup="from pendulum.parsing import parse_iso8601")
count, time_taken = timer.autorange()
timing = float(time_taken) / int(count)
print(f"{count}, {time_taken} => #{format_duration(timing)}")
```

### What should reviewers focus on?

I didn't bother supporting Python 2.7 in these changes. It's fairly easy to do so, but I noticed that pendulum [dropped support for Python < 3.7](https://github.com/sdispater/pendulum/pull/678), so I didn't bother.

I'm not sure why v2.1.2 performed so well in this test. I didn't look into it. 🤷 

## Pull Request Check List

- [ ] ~Added **tests** for changed code.~
    - Interface didn't change.
- [ ] ~Updated **documentation** for changed code.~
    - Interface didn't change. No publicly facing changes. Just performance.

### DIFF
--- a/pendulum/parsing/_iso8601.c
+++ b/pendulum/parsing/_iso8601.c
@@ -204,7 +204,7 @@ static int FixedOffset_init(FixedOffset *self, PyObject *args, PyObject *kwargs)
  * def utcoffset(self, dt):
  *     return timedelta(seconds=self.offset * 60)
  */
-static PyObject *FixedOffset_utcoffset(FixedOffset *self, PyObject *args) {
+static PyObject *FixedOffset_utcoffset(FixedOffset *self, PyObject *dt) {
     return PyDelta_FromDSU(0, self->offset, 0);
 }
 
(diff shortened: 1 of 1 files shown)


Secrus:
Thanks for contributing, but our C extensions are being rewritten into Rust. Superseded by #721

PR #680, opened 2022-12-01
Update mypy and fix issues

bryanforbes:
## Pull Request Check List

This upgrades `mypy` to 0.991 and fixes the errors associated with it. I've also removed the `mypy` check from `pre-commit` and added a type checking step during unit testing. This requires running `poetry install` twice to remove the type checking dependencies to ensure that there are no runtime dependencies on the type checking libraries (like `typing_extensions`). Another option would be to change the `pre-commit` config to use `mypy` via `poetry`, remove the third-party `pre-commit` PR hook, and add a Github action that invokes `pre-commit` with the proper `poetry install`. That can be done as a separate PR, though.


.github/workflows/tests.yml
```diff
@@ -64,8 +64,14 @@ jobs:
...
-      - name: Install dependencies
-        run: poetry install --only main --only test -vvv
+      - name: Install typing dependencies
+        run: poetry install --only main --only test --only typing -vvv
+
+      - name: Run type checking
+        run: poetry run mypy
+
+      - name: Install runtime and testing dependencies
+        run: poetry install --only main --only test --sync -vvv
```
Secrus, line 74: Not sure if we need to install project twice. We waste CI time for something that doesn't give us that much benefit

pyproject.toml
```diff
@@ -49,7 +49,9 @@ markdown-include = "^0.5.1"
 black = { version = "^22.6.0", markers = "implementation_name != 'pypy'" }
 isort = "^5.10.1"
 pre-commit = "^2.20.0"
-types-backports = "^0.1.3"
```
Secrus, line 52: Don't we need that for `zoneinfo`?

bryanforbes: No. `backports.zoneinfo` includes stub files. For some reason the type check isn't working on Windows and seems to be a [mypy bug](https://github.com/python/mypy/issues/14057#issuecomment-1334528626).

.github/workflows/tests.yml
bryanforbes: This line could be changed to `poetry install --only main --only test --sync --no-root -vvv` so that the build only occurs once

Secrus:
@bryanforbes could you update to the latest version of `mypy` and fix potential issues? I will have some time this weekend and want to merge this.

bryanforbes:
@Secrus this is ready to go

Secrus (approved):

PR #693, opened 2023-02-28
fix: `pendulum.parse` is not marked as exported

### LINKED ISSUE #595: pendulum.parse is not marked as exported (pyright type checker in VSCode)
- [x] I have searched the [documentation](https://pendulum.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

pyright type checker in VSCode reports a `reportPrivateImportUsage` error on following:

```
import pendulum
pendulum.parse('')
```

According to [pyright docs](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface) "Imported symbols are considered private by default." Which is the case of [pendulum.parse](https://github.com/sdispater/pendulum/blob/994e466e0fa620a769e70af28d75d1ede383be11/pendulum/__init__.py#L40).

It seems that simply changing the line to named export resolves the problem.

```patch
-from .parser import parse
+from .parser import parse as parse
```

This is the case for all imported symbols, but I came across it on `pendulum.parse`. I'm aware it's probably just pyright-specific issue, so I'm reporting it as Feature Request. I'm also willing to create a PR if you consider this a valid request.

cosmix:
This PR fixes #595, by re-exporting `parse()`. 

`parse()` should be re-exported so that the user can just call `pendulum.parse()` as documented, given that the module is marked with `py.typed`. 

No tests have been modified or added for this bugfix.

### DIFF
--- a/src/pendulum/__init__.py
+++ b/src/pendulum/__init__.py
@@ -29,7 +29,7 @@ from pendulum.helpers import set_locale
 from pendulum.helpers import week_ends_at
 from pendulum.helpers import week_starts_at
 from pendulum.interval import Interval
-from pendulum.parser import parse
+from pendulum.parser import parse as parse
 from pendulum.testing.traveller import Traveller
 from pendulum.time import Time
 from pendulum.tz import UTC


cosmix:
Note that the failing CI test has to do with the fact that the flake8 module that you're using seems to not like the import aliasing that _needs_ to be used to fix the issue addressed by this PR and that exports the `parse` symbol in `__init__.py`. 

For reference, I'm using the type checker rules listed on the pyright README [here](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface), but, to my understanding these are rules common to more type checkers for Python.

navmed (approved):

Secrus (approved):

Secrus:
Hi. Sorry for the delays, we are digging ourselves out of the pile of code debt. Could you please rebase your changes with the master branch so we can review and merge them?

PR #698, opened 2023-03-20
Add equality test

### LINKED ISSUE #502: Timezone classes should implement `__eq__`
Identical timezones don't compare equal, since equality is currently implemented via instance comparison.

```python
In [28]: from pendulum.tz.timezone import Timezone

In [29]: t1 = Timezone("UTC")

In [30]: t2 = Timezone("UTC")

In [31]: t1 == t2
Out[31]: False
```

This is most likely to come up when pickling `pendulum` instances - when unpickled the timezone instances will no longer match:

```python
In [44]: import pendulum

In [45]: import pickle

In [46]: t = pendulum.now()

In [47]: t2 = pickle.loads(pickle.dumps(t))

In [48]: t.tzinfo
Out[48]: Timezone('America/Chicago')

In [49]: t2.tzinfo
Out[49]: Timezone('America/Chicago')

In [50]: t.tzinfo == t2.tzinfo.  # I'd expect this to be true
Out[50]: False
```

This came up when using pendulum objects with `pandas` and `dask`. Some pandas methods validate the datetime objects input by checking that all provided timezones are identical. After deserializing the pendulum objects, the timezones no longer compared identical, even though they were equivalent.

---

It looks like there was an intent to cache instances with the same name here: 

https://github.com/sdispater/pendulum/blob/daa4b936daf3f4dfa7d211aa0ac1e9d82d5401d4/pendulum/tz/__init__.py#L34-L35

Since the unpickling code doesn't go through this path, the cache isn't used on load. Either the caching should be moved into the classes themselves, or `__eq__` should be implemented so multiple equivalent instances still compare equal.

malthe:
Closes #502.

Note that two timezones are equal only if their key (name) is the same string.

## Pull Request Check List

### DIFF
--- a/src/pendulum/tz/timezone.py
+++ b/src/pendulum/tz/timezone.py
@@ -7,6 +7,7 @@ import zoneinfo
 from abc import ABC
 from abc import abstractmethod
 from typing import TYPE_CHECKING
+from typing import Any
 from typing import TypeVar
 from typing import cast
 
@@ -65,6 +66,9 @@ class Timezone(zoneinfo.ZoneInfo, PendulumTimezone):
             return super().__new__(cls, key)  # type: ignore[call-arg]
         except zoneinfo.ZoneInfoNotFoundError:
             raise InvalidTimezone(key)
+
+    def __eq__(self, other: Any) -> bool:
+        return isinstance(other, PendulumTimezone) and self.key == other.key
 
     @property
     def name(self) -> str:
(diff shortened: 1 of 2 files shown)


pendulum/tz/timezone.py
```diff
@@ -58,6 +58,9 @@ def __new__(cls, key: str) -> Timezone:
         except zoneinfo.ZoneInfoNotFoundError:
             raise InvalidTimezone(key)
 
+    def __eq__(self, other: Any) -> bool:
```
Secrus, line 61:
```suggestion
    def __eq__(self, other: object) -> bool:
```

malthe: What's the difference in this case?

Secrus: Consistency. We already use `object` in other dunder methods and IIRC it's recommended.

PR #707, opened 2023-05-05
Allow subclassing DateTimes to work as expected

### LINKED ISSUE #203: pendulum.DateTime using hardcoded pendulum.datetime in methods
I found 4 methods in pendulum.datetime that return a new pendulum.datetime object. The pendulum.datetime class is hardcoded in these, which could cause some inconstistencies when subclassing:

```
import pendulum

class MyDateTime(pendulum.DateTime):
    pass

MyDateTime.now()
```

The output is an instance of pendulum.DateTime while I would expect it to be an instance of MyDateTime. 

This makes extending the library a bit of a hassle.

gazpachoking:
Currently if subclassing DateTime, there are many methods which would return a plain DateTime type, rather than the subclassed type.

I made a new `DateTime._instance` classmethod so as not to change the public interface, which is used by `pendulum.instance` as well as internally in the DateTime class.

Fixes #203

## Pull Request Check List

These changes don't affect using the library directly, only when subclassed externally. Existing tests should ensure internal behavior hasn't changed.


gazpachoking:
Any way to get some eyes on this? I'd love to get this working.

sdispater:
Thanks for this PR. This looks good however I integrated this change in #732 which includes improvements to `instance()`. I co-authored you in this PR as well 🙂

gazpachoking:
Adding the instance method to the class was my preferred method as well, but I didn't want to change the API without blessing. Even more useful that it's public, thanks so much!

PR #713, opened 2023-06-14
fix: handle DST folded dates with start/end_of

nekonyuu:
## The bug

While implementing pendulum usage where I work (switching from buggy `arrow` DST handling), I've found a bug around start_of/end_of handling of dates, when the passed date is exactly in the ambiguous hour period (going out of DST).

For example : 

```python
>>> d = pendulum.datetime(2013, 10, 27, 0, 59, 59, tz="UTC").in_timezone("Europ
>>> d.isoformat()
'2013-10-27T02:59:59+02:00'
>>> d.start_of("hour").isoformat()
'2013-10-27T02:00:00+01:00'
```

Here, the expected behavior would be that I still get the offset from the current hour (as I didn't transition).
The end_of method suffer of the same bug.

## Fix proposal

I'm not that familiar with Python and its internal about fold, but reading the PEP, I've found that simply porting the fold field from the original date is enough, as we were missing the information that the date either didn't happen yet, or did.

## Pull Request Check List


sdispater (approved):

sdispater:
Thanks!

Issue #587, opened 2021-12-10
week_of_month in pendulum_parse for Timestamp '2021-01-15 14:00:00' returns -50


vidyasagar86:
week_of_month in pendulum_parse for Timestamp '2021-01-15 14:00:00' returns -50

Sample code:
import pendulum

x='2021-01-15 14:00:00'
dt=pendulum.parse(x)
print(dt.week_of_month)

Output result -50

anvit25:
I looked a bit into it. Seems like Jan 1 2021 is defined as 5th weekday of the 53rd week of 2020

dmateusp:
This is an issue in 2023 as well

```python
>>> pendulum.instance(datetime(2023, 1, 2, 0, 0)).week_of_month
-50
```

using `pendulum==2.1.2`

This library is used in Airflow so I imagine more people will encounter it

pylead:
Using `pendulum==2.1.2` confirm, the issue presents.
```python
>>> import pendulum
>>> pendulum.DateTime(2023, 1, 2).week_of_month
-50
```

PR #718, opened 2023-06-27
Fix: week_of_month returns negative values

ulasozguler:
## Pull Request Check List

## Description

Resolves https://github.com/sdispater/pendulum/issues/587

## Changes

Using day of month instead of `week_of_year` to calculate `week_of_month`.

### DIFF
--- a/pendulum/date.py
+++ b/pendulum/date.py
@@ -27,6 +27,7 @@ from pendulum.constants import YEARS_PER_CENTURY
 from pendulum.constants import YEARS_PER_DECADE
 from pendulum.exceptions import PendulumException
 from pendulum.helpers import add_duration
+from pendulum.helpers import days_of_week
 from pendulum.interval import Interval
 from pendulum.mixins.default import FormattableMixin
 
@@ -83,8 +84,11 @@ class Date(FormattableMixin, date):
     @property
     def week_of_month(self) -> int:
         first_day_of_month = self.replace(day=1)
-
-        return self.week_of_year - first_day_of_month.week_of_year + 1
+        week_days = days_of_week()
+        first_day_index = week_days.index(first_day_of_month.day_of_week)
+        days_in_first_week = pendulum.DAYS_PER_WEEK - first_day_index
+        days_after_first_week = self.day - days_in_first_week
+        return math.ceil(days_after_first_week / pendulum.DAYS_PER_WEEK) + 1
 
     @property
     def age(self) -> int:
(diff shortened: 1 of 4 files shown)


tests/date/test_getters.py
```diff
@@ -61,6 +61,9 @@ def test_week_of_month():
     assert pendulum.date(2020, 1, 1).week_of_month == 1
     assert pendulum.date(2020, 1, 7).week_of_month == 2
     assert pendulum.date(2020, 1, 14).week_of_month == 3
+    assert pendulum.date(2023, 1, 1).week_of_month == 1
```
sdispater, line 64: This is not correct, January 1st 2023 belongs to the last week of December of 2022. Since this is the edge case, it should return 0.

tests/date/test_getters.py
```diff
@@ -61,6 +61,9 @@ def test_week_of_month():
...
+    assert pendulum.date(2023, 1, 2).week_of_month == 2
+    assert pendulum.date(2023, 1, 31).week_of_month == 6
```
sdispater, line 66: This is not correct, from an ISO standpoint it's not possible to have a 6th week, it should be 5.

pendulum/date.py
```diff
@@ -83,8 +84,11 @@ def days_in_month(self) -> int:
     @property
     def week_of_month(self) -> int:
         first_day_of_month = self.replace(day=1)
-
-        return self.week_of_year - first_day_of_month.week_of_year + 1
+        week_days = days_of_week()
+        first_day_index = week_days.index(first_day_of_month.day_of_week)
+        days_in_first_week = pendulum.DAYS_PER_WEEK - first_day_index
+        days_after_first_week = self.day - days_in_first_week
+        return math.ceil(days_after_first_week / pendulum.DAYS_PER_WEEK) + 1
```
sdispater, line 91:
```suggestion
        return int(
            math.ceil((self.day + first_day_of_month.day_of_week - 1) / DAYS_PER_WEEK)
        )
```

This should give us what we want while taking care of the edge cases that needs fixing.

tests/date/test_getters.py
```diff
@@ -61,6 +61,9 @@ def test_week_of_month():
     assert pendulum.date(2020, 1, 1).week_of_month == 1
     assert pendulum.date(2020, 1, 7).week_of_month == 2
     assert pendulum.date(2020, 1, 14).week_of_month == 3
+    assert pendulum.date(2023, 1, 1).week_of_month == 1
+    assert pendulum.date(2023, 1, 2).week_of_month == 2
```
sdispater, line 65:
```suggestion
    assert pendulum.date(2023, 1, 2).week_of_month == 1
```

sdispater (changes requested):

PR #721, opened 2023-07-18
Rewrite extensions in Rust

sdispater:
This PR is the first introduction to Rust in the codebase.

It is a complete rewrite of existing C extensions and paves the way to more speed improvements by rewriting other parts of the codebase in the future.

Note that if there is still a pure Python version of these extensions and will always be for the foreseeable future.

Be aware that this PR introduces small backwards  incompatibilities when parsing ISO8601 strings:

- Basic time representation now requires a leading `T`, for instance `T201205`
- Fractional hours and minutes are no longer supported. It was technically supported before the behavior was incorrect so support has been removed for now but will be reintroduced in the future.

## Pull Request Check List


Secrus (changes requested):
Some issues here and there, but overall, nice job. I am no Rust expert, so I just looked through it trusting that the test suite checks if it works correctly.

.github/workflows/tests.yml
```diff
@@ -13,6 +13,20 @@ on:
...
 jobs:
+  Linting:
+    name: Linting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: "Install pre-commit"
```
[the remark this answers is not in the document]
sdispater: Unfortunately, the clippy hook I want is not supported in pre-commit CI: when Cargo tries to download clippy there is a connection error.

So, for now, the only workaround I found is to not use pre-commit CI.

Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "pendulum"
+version = "3.0.0-alpha-1"
+edition = "2021"
+
+[lib]
+name = "_pendulum"
```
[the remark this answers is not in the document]
sdispater: I don't want users to import directly from the extension, that's why I chose this name. I might change it in the future if. more features are rewritten in Rust.

Makefile
```diff
@@ -51,3 +51,71 @@ build_wheels_i686:
 # run tests against all supported python versions
 tox:
 	@tox
+
+
+lint-rust:
+	cargo fmt --version
+	cargo fmt --all -- --check
+	cargo clippy --version
+	cargo clippy --tests -- \
```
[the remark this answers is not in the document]
sdispater: `clippy` supports a config file but those settings cannot be put in it.

I can remove the version printing.

pendulum/helpers.py
```diff
@@ -31,23 +31,21 @@
     if not with_extensions or struct.calcsize("P") == 4:
         raise ImportError()
 
-    from pendulum._extensions._helpers import PreciseDiff
-    from pendulum._extensions._helpers import days_in_year
-    from pendulum._extensions._helpers import is_leap
-    from pendulum._extensions._helpers import is_long_year
-    from pendulum._extensions._helpers import local_time
-    from pendulum._extensions._helpers import precise_diff
-    from pendulum._extensions._helpers import timestamp
```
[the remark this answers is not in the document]
sdispater: `timetamp()` is no longer used. It's a leftover of a previous pendulum version.

rust/helpers.rs
```diff
@@ -0,0 +1,122 @@
+use crate::constants::{
+    DAYS_PER_L_YEAR, DAYS_PER_N_YEAR, DAY_OF_WEEK_TABLE, EPOCH_YEAR, MONTHS_OFFSETS,
+    SECS_PER_100_YEARS, SECS_PER_400_YEARS, SECS_PER_4_YEARS, SECS_PER_DAY, SECS_PER_HOUR,
+    SECS_PER_MIN, SECS_PER_YEAR, TM_DECEMBER, TM_JANUARY,
+};
+
+fn p(year: i32) -> i32 {
```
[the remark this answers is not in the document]
sdispater: I don't have a good name for it, unfortunately, it's just a way to determine the week day of January 1st of the given year: `weekday = p(year) % 7`.

I don't think it matters too much, it's purely internal and is just a part of known date algorithm.

build.py
```diff
@@ -1,19 +1,36 @@
...
+        cargo_args = shlex.split(os.getenv("MATURIN_BUILD_ARGS", ""))
+
+    maturin("build", "-r", *cargo_args)
+
+    wheel = list(wheels_dir.glob("*.whl"))[0]
+    with zipfile.ZipFile(wheel.as_posix()) as whl:
+        whl.extractall(wheels_dir.as_posix())
+
+        for extension in wheels_dir.rglob("**/*.so"):
+            shutil.copyfile(extension, Path(__file__).parent.joinpath(extension.name))
```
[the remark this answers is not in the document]
sdispater: I added a comment

pyproject.toml
```diff
@@ -20,8 +20,13 @@ include = [
     # C extensions must be included in the wheel distributions
     { path = "pendulum/_extensions/*.so", format = "wheel" },
```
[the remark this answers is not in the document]
sdispater: Done!

pyproject.toml
```diff
@@ -219,5 +224,5 @@ omit = [
 ]
 
 [build-system]
-requires = ["poetry-core>=1.1.0a6", "meson", "ninja"]
+requires = ["poetry-core>=1.1.0a6", "maturin>=1,<2"]
```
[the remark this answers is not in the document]
sdispater: Done!

pendulum/parser.py
```diff
@@ -13,14 +14,13 @@
 if t.TYPE_CHECKING:
     from pendulum.date import Date
     from pendulum.datetime import DateTime
-    from pendulum.duration import Duration
     from pendulum.interval import Interval
     from pendulum.time import Time
 
 try:
-    from pendulum.parsing._iso8601 import Duration as CDuration
+    from _pendulum import Duration as CDuration
```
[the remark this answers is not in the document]
sdispater: Changed to `RustDuration`

rust/parsing.rs
```diff
@@ -0,0 +1,945 @@
...
+                return Err(self.parse_error(format!(
+                    "Unexpected end of string while parsing \"{}\". Expected {} more character{}",
+                    field_name,
+                    length - i,
+                    if (length - i) != 1 { "s" } else { "" }
+                )));
+            }
+
+            if self.current >= '0' && self.current <= '9' {
+                value = 10 * value + self.current.to_digit(10).unwrap();
```
Maneren, line 192: This can imo be better expressed using `if-let` and `to_digit` directly. `.unwrap` is usually unidiomatic, even if it can't fail
```suggestion
            if let Some(digit) = self.current.to_digit(10) {
                value = 10 * value + digit;
```
also on lines 480 and 530

rust/parsing.rs
```diff
@@ -0,0 +1,945 @@
...
+                            } else {
+                                break;
+                            }
+
+                            self.inc();
+                            i += 1;
+                        }
+
+                        // Drop extraneous digits
+                        while self.current >= '0' && self.current <= '9' {
```
Maneren, line 545: std has a helper method for this
```suggestion
                        while self.current.is_ascii_digit() {
```

rust/parsing.rs
```diff
@@ -0,0 +1,945 @@
...
+                            (datetime.month * 10 + self.parse_integer(1, "ordinal day")?) as i32;
+
+                        match self.ordinal_to_ymd(datetime.year, ordinal_day, false) {
+                            Ok((year, month, day)) => {
+                                datetime.year = year;
+                                datetime.month = month;
+                                datetime.day = day;
+                            }
+                            Err(error) => return Err(error),
+                        }
```
Maneren, line 332: this can be simplified using the `?` operator
```suggestion
                        let (year, month, day) =
                            self.ordinal_to_ymd(datetime.year, ordinal_day, true)?;

                        datetime.year = year;
                        datetime.month = month;
                        datetime.day = day;
```

rust/parsing.rs
```diff
@@ -0,0 +1,945 @@
...
+                self.current = ch;
+                true
+            }
+            None => {
+                self.idx = self.src.len();
+                self.current = '\0';
+                false
+            }
+        }
+    }
```
Maneren, line 140: Returning `Option<char>` allows us to simplify code in some places (line 848 and 856), while not affecting current code as the `bool` isn't read anywhere.
```suggestion
    /// Returns the next character if it exists.
    fn inc(&mut self) -> Option<char> {
        if let Some((i, ch)) = self.chars.next() {
            self.idx = i;
            self.current = ch;
            Some(ch)
        } else {
            self.idx = self.src.len();
            self.current = '\0';
            None
        }
    }
```

rust/parsing.rs
```diff
@@ -0,0 +1,945 @@
...
+            self.inc();
+
+            match self.current {
+                c if c.is_ascii_digit() => {
+                    value *= 10;
+                    value += c.to_digit(10).unwrap();
+                }
+                _ => return Ok(value),
+            }
+        }
```
Maneren, line 866: Use the `self.inc()` change from line 127 to simplify the looping condition.
```suggestion
        while let Some(digit) = self.inc().and_then(|ch| ch.to_digit(10)) {
            value *= 10;
            value += digit;
        }

        Ok(value)
```

rust/parsing.rs
```diff
@@ -0,0 +1,945 @@
...
+    }
+
+    fn parse_duration_number(&mut self) -> Result<u32, ParseError> {
+        let mut value = match self.current {
+            c if c.is_ascii_digit() => c.to_digit(10).unwrap(),
+            _ => {
+                return Err(self.parse_error("Invalid number in duration".to_string()));
+            }
+        };
```
Maneren, line 854: This can be simplified using `let-else` and the principle from line 191.
```suggestion
        let Some(mut value) = self.current.to_digit(10) else {
            return Err(self.parse_error("Invalid number in duration".to_string()));
        };
```

rust/parsing.rs
```diff
@@ -0,0 +1,945 @@
...
+                        decimal *= 10.0;
+                        decimal += f64::from(c.to_digit(10).unwrap());
+                        denominator *= 10.0;
+                    }
+                    _ => return Ok((value, Some(decimal / denominator))),
+                }
+            }
+        } else {
+            Ok((value, None))
+        }
```
Maneren, line 845: Refactor the duplicated return value.
Plus same as line 856.
```suggestion
        let fraction = if self.current == '.' || self.current == ',' {
            let mut decimal = 0_f64;
            let mut denominator = 1_f64;

            while let Some(digit) = self.inc().and_then(|ch| ch.to_digit(10)) {
                decimal *= 10.0;
                decimal += f64::from(digit);
                denominator *= 10.0;
            }

            Some(decimal / denominator)
        } else {
            None
        };

        Ok((value, fraction))
```
With the `matches!` change you can get rid of the else case using `.then`, but that's highly opinionated so feel free to reject it.
```suggestion
        let fraction = matches!(self.current, '.' | ',').then(|| {
            let mut decimal = 0_f64;
            let mut denominator = 1_f64;

            while let Some(digit) = self.inc().and_then(|ch| ch.to_digit(10)) {
                decimal *= 10.0;
                decimal += f64::from(digit);
                denominator *= 10.0;
            }

            decimal / denominator
        });

        Ok((value, fraction))
```

Maneren (changes requested):
Great initiative, seeing C rewritten to Rust always makes me happy. I suggested some minor refactorings, partially suggested by `clippy::pedantic`, but otherwise good job. There are also some notable remnants of C (eg. the null char) but those require more complex changes and aren't really that necessary.

sdispater:
Thanks for the thorough review @Maneren! I applied all of your suggestions.

Maneren:
Happy to help

pyproject.toml
```diff
@@ -17,11 +17,13 @@ packages = [
 include = [
     { path = "meson.build", format = "sdist" },
```
Secrus, line 18: Meson script won't be needed anymore.

build.py
```diff
@@ -1,19 +1,39 @@
...
 
 
 def _build():
     build_dir = Path(__file__).parent.joinpath("build")
     build_dir.mkdir(parents=True, exist_ok=True)
 
-    meson("setup", build_dir.as_posix())
-    meson("compile", "-C", build_dir.as_posix())
-    meson("install", "-C", build_dir.as_posix())
+    wheels_dir = Path(__file__).parent.joinpath("target/wheels")
```
Secrus, line 18: I wonder if we should add cleanup/deletion of maturin-built wheels.

build.py
```diff
@@ -22,11 +42,12 @@ def build(setup_kwargs):
     """
     try:
         _build()
-    except Exception:
+    except Exception as e:
         print(
             "  Unable to build C extensions, "
```
Secrus, line 47:
```suggestion
            "  Unable to build Rust extensions, "
```

Secrus (changes requested):
Few extra nitpicks

pyproject.toml
Secrus: You should also remove `build` group from dependencies

sdispater: Done!

build.py
sdispater: I added the cleanup at the end of the build step

build.py
sdispater: Done!

Secrus (approved):
LGTM, great job!

PR #521, opened 2020-11-19
Add wheel support for aarch64

odidev:
Added wheel support for aarch64.
Related to #517  .

Issue #517, opened 2020-11-02
Wheel support for linux aarch64


odidev:
**Summary**
Installing pendulum on aarch64 via pip using command "pip3 install pendulum" tries to build wheel from source code.

**Problem description**
pendulum doesn't have wheel for aarch64 on PyPI repository. So, while installing pendulum via pip on aarch64, pip builds wheel for same resulting in it takes more time to install pendulum. Making wheel available for aarch64 will benefit aarch64 users by minimizing pendulum installation time.

**Expected Output**
Pip should be able to download pendulum wheel from PyPI repository rather than building it from source code.

@pendulum-team, please let me know if I can help you building wheel/uploading to PyPI repository. I am curious to make pendulum wheel available for aarch64. It will be a great opportunity for me to work with you.

Secrus:
Pendulum 3.0 was released, along with wheels for aarch64

PR #740, opened 2023-08-17
Fix the release workflow

sdispater:
For the release workflow to work, we need to install the Rust toolchain first on some platforms and we need to use Python 3 to fix the wheels for MacOS arm wheels.

## Pull Request Check List

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,6 +18,16 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Set up rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: stable
+
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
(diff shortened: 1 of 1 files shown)


Secrus:
what about linux aarch64 wheels? maybe we could copy `coverage` setup? https://github.com/nedbat/coveragepy/blob/c2b238e86cade2e6163b24656a7d0549f4508e8f/.github/workflows/kit.yml#L101

sdispater:
@Secrus We can always add the `aarch64` architecture but I don't know if we will need the same fixes as the ones needed for MacOS due to https://github.com/python-poetry/poetry/issues/7107.

Secrus:
@sdispater I just thought it's worth adding and checking since you are already working on the release pipeline. Especially since there was a PR/request for that architecture for some time now: #517 #521

edgarrmondragon:
Since this project seems to use maturin as a build, did you folks consider using the auto-generated CI build workflow?

```
maturin generate-ci github > .github/workflows/CI.yml
```

PR #741, opened 2023-08-19
Use re.fullmatch to match whole strings

eumiro:


Secrus:
Hi. Sorry for the delays, we are digging ourselves out of the pile of code debt. Could you please rebase your changes with the master branch so we can review and merge them?

Secrus (approved):

PR #752, opened 2023-09-26
Kyiv not Kiev

zabop:
Correction to use the official spelling. Microsoft is also using the Ukrainian, rather than the Russian name: https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/default-time-zones?view=windows-11. (The city is the capital of Ukraine, that's why.)


Secrus (approved):

Secrus:
Could you rebase? Seems like a simple fix that would be easy to merge

PR #778, opened 2023-12-16
Make the test helpers opt-in via a test extra

sdispater:
Since the testing helpers are only necessary when used in tests, this PR makes them optional via a `test` extra.

## Pull Request Check List


pyproject.toml
```diff
@@ -18,7 +18,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 
-dependencies =[
```
edgarrmondragon, line 21: Should `time-machine` be moved to `optional-dependencies`?

PR #786, opened 2023-12-21
Restore period function with deprecation warning

mathieulongtin:
## Pull Request Check List

### DIFF
--- a/src/pendulum/__init__.py
+++ b/src/pendulum/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import datetime as _datetime
+import warnings
 
 from typing import Union
 from typing import cast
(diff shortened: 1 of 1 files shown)


Secrus:
Why? This was an intentional breaking change followed by a major version bump. I don't see any reason to bring this back.

mathieulongtin:
> Why? This was an intentional breaking change followed by a major version bump. I don't see any reason to bring this back.

We’ve been using pendulum for five year or so. It’s solid, it works, it fixes gaps in the datetime library. 

But changing a function name with no depreciation warning after years with the old name is not cool. Major version change or not. Now we have to worry about dependency mismatches, and wonder if other naming changes will be sprung on us.

PR #803, opened 2024-02-03
Avoid usage of deprecated `datetime.datetime.utcfromtimestamp` in `pendulum.from_timestamp`

Taragolis:
## Pull Request Check List

`datetime.datetime.utcfromtimestamp()` is deprecated into the [Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#deprecated) as result current implementation raise warning message

```console
/usr/local/lib/python3.12/site-packages/pendulum/__init__.py:295 DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC:
```

### DIFF
--- a/src/pendulum/__init__.py
+++ b/src/pendulum/__init__.py
@@ -292,11 +292,7 @@ def from_timestamp(timestamp: int | float, tz: str | Timezone = UTC) -> DateTime
     """
     Create a DateTime instance from a timestamp.
     """
-    dt = _datetime.datetime.utcfromtimestamp(timestamp)
-
-    dt = datetime(
-        dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second, dt.microsecond
-    )
+    dt = DateTime.fromtimestamp(timestamp, tz=UTC)
 
     if tz is not UTC or tz != "UTC":
         dt = dt.in_timezone(tz)


src/pendulum/__init__.py
```diff
@@ -292,11 +292,7 @@ def from_timestamp(timestamp: int | float, tz: str | Timezone = UTC) -> DateTime
     """
     Create a DateTime instance from a timestamp.
     """
-    dt = _datetime.datetime.utcfromtimestamp(timestamp)
-
-    dt = datetime(
-        dt.year, dt.month, dt.day, dt.hour, dt.minute, dt.second, dt.microsecond
-    )
+    dt = DateTime.fromtimestamp(timestamp, tz=UTC)
```
SinTh0r4s, line 295: I think the current suggestion is recursive and does not use the correct keyword:
```suggestion
    dt = _datetime.datetime.fromtimestamp(timestamp, tzinfo=UTC)
```

For the develop branch, the change must be done in `pendulum/pendulum.py:491`
```python
dt = datetime.datetime.utcfromtimestamp(timestamp).replace(tzinfo=UTC)
```
to
```python
dt = datetime.datetime.fromtimestamp(timestamp, tzinfo=UTC)
```

hozn:
Also interested in this fix; is there help needed to get this through?  (I see a failing test, but unsure if that's accurate as this is pretty old.)

Issue #784, opened 2023-12-19
Unable to build pendulum for 32bit


hesstobi:
- [ x] I am on the [latest](https://github.com/sdispater/pendulum/releases/latest) Pendulum version.
- [x ] I have searched the [issues](https://github.com/sdispater/pendulum/issues) of this repo and believe that this is not a duplicate.

- **OS version and name**: Linux 32bit
- **Pendulum version**: 3.0.0

## Issue
Trying to build pendulum for a 32bit system fails:

```
-> src/helpers.rs:59:20
         |
      59 |         seconds += ((146_097 - 10957) * SECS_PER_DAY as usize) as isize;
         |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `135140_usize * 86400_usize`, which would overflow
         |
         = note: `#[deny(arithmetic_overflow)]` on by default
```

usize is only a 32 uint on 32 bit systems, thus this is not working.

Secrus:
Unless you are up to making the change required for this to work on 32 bit system, I doubt it will change.

hesstobi:
This was my concern. 

As far I understand it would work, if the seconds variable in is changed to i64.

I can try to start with rust programming and prepare a PR for that.

gazpachoking:
Is there a way to force the install to fall back to the pure python implementation?

edgarrmondragon:
> Is there a way to force the install to fall back to the pure python implementation?
> 
> EDIT: Or could a pure python wheel be published to pypi to fall back on?

@gazpachoking maybe the `PENDULUM_EXTENSIONS=0` env var?

gazpachoking:
I tried that env var. It seems that only disables the extensions at run
time. I don't see a way to install without building the rust extensions at
all.

gazpachoking:
We added pendulum as a dependency in our project and are now having multiple reports of people who can't install on their platform because either it's 32 bit, or they don't have rust tooling and there isn't a prebuilt wheel for their platform. I really hope official way is added such that pendulum can fall back to the pure python implementation in those cases.

Secrus:
> We added pendulum as a dependency in our project and are now having multiple reports of people who can't install on their platform because either it's 32 bit, or they don't have rust tooling and there isn't a prebuilt wheel for their platform. I really hope official way is added such that pendulum can fall back to the pure python implementation in those cases.

[...]

PR #842, opened 2024-09-05
rust: Use i64 for internal unix timestamps

bdrung:
pendulum 3.0.0 fails to build on 32-bit armhf:

```
error: this arithmetic operation will overflow
  --> src/helpers.rs:59:20
   |
59 | seconds += ((146_097 - 10957) * SECS_PER_DAY as usize) as isize;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `135140_usize * 86400_usize`, which would overflow
   |
   = note: `#[deny(arithmetic_overflow)]` on by default
```

`(146_097 - 10957) * SECS_PER_DAY` equals 11,676,096,000 which does not fit into 32 bit integers.

Use i64 for the seconds variable while handling with the timestamp. Only convert in to `usize` once the timestamp is split into its components.

Fixes https://github.com/sdispater/pendulum/issues/784

## Pull Request Check List

### DIFF
--- a/rust/src/helpers.rs
+++ b/rust/src/helpers.rs
@@ -49,57 +49,57 @@ pub fn local_time(
     microsecond: usize,
 ) -> (usize, usize, usize, usize, usize, usize, usize) {
     let mut year: usize = EPOCH_YEAR as usize;
-    let mut seconds: isize = unix_time.floor() as isize;
+    let mut seconds: i64 = unix_time.floor() as i64;
 
     // Shift to a base year that is 400-year aligned.
     if seconds >= 0 {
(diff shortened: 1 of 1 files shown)


Secrus:
Thank you for your patch. Could you please fix the issues highlighted by the CI? You should be able to reproduce this by running `make lint-rust`. Thank you

Secrus:
Please rebase your changes with master to fix the failing pipeline

bdrung:
Rebased on master.

Secrus (approved):

Secrus:
Thank you for the fix

PR #843, opened 2024-09-11
Fix parsing invalid intervals

wiget:
Parsing a string with a slash at the start or the end raised IndexError instead of ParseError.

## Pull Request Check List

### DIFF
--- a/src/pendulum/parsing/__init__.py
+++ b/src/pendulum/parsing/__init__.py
@@ -214,11 +214,11 @@ def _parse_iso8601_interval(text: str) -> _Interval:
     first, last = text.split("/")
     start = end = duration = None
 
-    if first[0] == "P":
+    if first[0:1] == "P":
         # duration/end
         duration = parse_iso8601(first)
         end = parse_iso8601(last)
-    elif last[0] == "P":
+    elif last[0:1] == "P":
         # start/duration
         start = parse_iso8601(first)
         duration = parse_iso8601(last)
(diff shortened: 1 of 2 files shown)


wiget:
@sdispater would you mind reviewing this PR?

src/pendulum/parsing/__init__.py
```diff
@@ -214,11 +214,11 @@ def _parse_iso8601_interval(text: str) -> _Interval:
     first, last = text.split("/")
     start = end = duration = None
 
-    if first[0] == "P":
+    if first[0:1] == "P":
```
Secrus, line 217: How does that change anything? AFAIK `[0]` is the same as `[0:1]`.

bswck: It's not the same. `[0:1]`, which could also be written as `[:1]`, is a lenient version of `[0]` -- it prevents from raising an `IndexError` on an empty string and instead evaluates to `''` in that case.

bswck:
If I could suggest something, it would be just using `[:1]` instead of `[0:1]`, as the former simply works and is less unnecessarily explicit.

Similarly to how you would probably prefer reading `range(10)` than `range(0, 10)`.

wiget:
@bswck You are right. Suggestion applied.

bswck (approved):

wiget:
@Secrus would you mind reviewing this PR?

Secrus (approved):

wiget:
Thanks for the review.

What is required to get this PR merged?

Secrus:
@wiget please rebase your branch with master so the pipelines run. Right now it looks like GH has some issues here.

wiget:
All checks passed. Would you mind merging this PR, @Secrus ?

bennetrr:
Is there a release planned for this?

PR #852, opened 2024-10-28
fix(parsing): handle cases where string starts/ends with a slash

temporaer:


temporaer:
abandoning in favor of #843

Issue #844, opened 2024-09-11
3.13 support


raphaelauv:
- [x] I have searched the [issues](https://github.com/sdispater/pendulum/issues) of this repo and believe that this is not a duplicate.
- [x] I have searched the [documentation](https://pendulum.eustace.io/docs/) and believe that my question is not covered.

## Feature Request

Hi, an official support of python 3.13, thanks

Secrus:
Is Pendulum incompatible right now?

M5oul:
There are no Python 3.13 wheel provided. That’s the least I can say.
Releasing a new version on PyPI would provide a 3.13 wheel I guess.

ericfrederich:
3.13 is released, hopefully a 3.13 wheel can be provided soon.  I don't want to install Rust to use Python ;-)

ollz272:
@sdispater Is it possible to get this in, please?

M5oul:
May be @Secrus?

edgarrmondragon:
@ericfrederich @ollz272 @M5oul see https://github.com/sdispater/pendulum/issues/590#issuecomment-2405924505

ericfrederich:
Some slightly bad news here.  While the tests are passing with Python 3.12 the tests do not pass on 3.13.

Python 3.13 [removed `uuid._load_system_functions`](https://github.com/python/cpython/pull/115934) which is called by `time-machine` which pendulum depends on.

Two ways of fixing this.

1: `poetry update time-machine` which updates time-machine from 2.13.0 -> 2.15.0 [where the call is gated based on sys.version_info](https://github.com/adamchainz/time-machine/pull/425).
2: bump minimum version to 3.9 which would update time-machine to 2.16.0 where they also dropped 3.8 support and [the call is removed](https://github.com/adamchainz/time-machine/pull/484)

rudolfix:
tbh. it is hard to say why `time-machine` is in the dependencies. its only use is for tests, still this stuff is import into main `__init__`. or am I missing something?

svrooij:
Are you still depending upon pendulum (the python library that requires rust to install), here is the PR how I removed the dependency upon it for Kiota.

It was quite easy since most functionality we were using is now supported by by Python itself. Actually the code we used is replaced by this code https://github.com/microsoft/kiota-python/blob/e9840dfd9acf75292e5c1cff3277a026a9a49140/packages/abstractions/kiota_abstractions/date_utils.py

[...]

PR #863, opened 2024-12-04
Drop Py3.8 support to add potential Py3.13 support

M5oul:
at least from the dependencies point of view.
Bumping dependencies, which dropped Py3.8 supports, also adds Py3.13 support.

(#844)


src/pendulum/utils/_zoneinfo.py
```diff
@@ -1,74 +1,12 @@
 from __future__ import annotations
 
-import sys
+from zoneinfo import TZPATH
+from zoneinfo import InvalidTZPathWarning
+from zoneinfo import ZoneInfo
+from zoneinfo import ZoneInfoNotFoundError
+from zoneinfo import available_timezones
+from zoneinfo import reset_tzpath
```
Secrus, line 8: This shouldn't be needed anymore, all uses of `pendulum.utils._zoneinfo` should be replaced with direct usage of zoneinfo and this module should be removed.

Secrus (commented):
One issue for now, let's see what CI says

src/pendulum/utils/_zoneinfo.py
M5oul: Done!
Thanks for triggering codspeed!
Regarding the performances regression, I am not sure if this is acceptable.
There is a loss of "performance" according to codspeed on [v2.0.0 release](https://github.com/CodSpeedHQ/pytest-codspeed/releases/tag/v2.0.0).

Secrus:
I am no expert on Rust/PyO3, but are the changes to the code needed? There is a huge performance regression. It's unmergable in that state.

M5oul:
I cherry-picked last commit from https://github.com/dlt-hub/pendulum/, otherwise, we get `rust-lint` errors that can be found bellow. I can’t say if the Rust changes are any good. I can’t say the which of rust changes or `codspeed` bump changed the performances.
```
error: non-local `impl` definition, `impl` blocks should be written at the same level as their item
  --> src/python/types/duration.rs:23:1
   |
23 | #[pymethods]
   | -^^^^^^^^^^^
   | |
   | `PyClassImplCollector` is not local
   | `PyClassNewTextSignature` is not local
   | move the `impl` block outside of this function `trampoline` and up 4 bodies
24 | impl Duration {
   |      --------
   |      |
   |      `Duration` is not local
   |      `Duration` is not local
   |
   = note: the attribute macro `pymethods` defines the non-local `impl`, and may need to be changed
   = note: the attribute macro `pymethods` may come from an old version of the `pyo3_macros` crate, try updating your dependency with `cargo update -p pyo3_macros`
   = note: an `impl` is never scoped, even when it is nested inside an item, as it may impact type checking outside of that item, which can be the case if neither the trait or the self type are at the same nesting level as the `impl`
   = note: `-D non-local-definitions` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(non_local_definitions)]`
   = note: this error originates in the attribute macro `pymethods` (in Nightly builds, run with -Z macro-backtrace for more info)

error: non-local `impl` definition, `impl` blocks should be written at the same level as their item
  --> src/python/types/precise_diff.rs:23:1
   |
23 | #[pymethods]
   | -^^^^^^^^^^^
   | |
   | `PyClassImplCollector` is not local
   | `PyClassNewTextSignature` is not local
   | move the `impl` block outside of this function `trampoline` and up 4 bodies
24 | impl PreciseDiff {
   |      -----------
   |      |
   |      `PreciseDiff` is not local
   |      `PreciseDiff` is not local
   |
   = note: the attribute macro `pymethods` defines the non-local `impl`, and may need to be changed
   = note: the attribute macro `pymethods` may come from an old version of the `pyo3_macros` crate, try updating your dependency with `cargo update -p pyo3_macros`
   = note: an `impl` is never scoped, even when it is nested inside an item, as it may impact type checking outside of that item, which can be the case if neither the trait or the self type are at the same nesting level as the `impl`
   = note: this error originates in the attribute macro `pymethods` (in Nightly builds, run with -Z macro-backtrace for more info)

error: non-local `impl` definition, `impl` blocks should be written at the same level as their item
  --> src/python/types/timezone.rs:11:1
   |
11 | #[pymethods]
   | -^^^^^^^^^^^
   | |
   | `PyClassImplCollector` is not local
   | `PyClassNewTextSignature` is not local
   | move the `impl` block outside of this function `trampoline` and up 4 bodies
12 | impl FixedTimezone {
   |      -------------
   |      |
   |      `FixedTimezone` is not local
   |      `FixedTimezone` is not local
   |
   = note: the attribute macro `pymethods` defines the non-local `impl`, and may need to be changed
   = note: the attribute macro `pymethods` may come from an old version of the `pyo3_macros` crate, try updating your dependency with `cargo update -p pyo3_macros`
   = note: an `impl` is never scoped, even when it is nested inside an item, as it may impact type checking outside of that item, which can be the case if neither the trait or the self type are at the same nesting level as the `impl`
   = note: this error originates in the attribute macro `pymethods` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `_pendulum` (lib test) due to 3 previous errors
make: *** [Makefile:13: lint-rust] Error 101
```

M5oul:
As said is codspeed message:

> #### 🎉 Hooray! `pytest-codspeed` just leveled up to 3.0.0!
> 
> > A heads-up, this is a **breaking change** and it might affect your current performance baseline a bit. But here's the exciting part - it's packed with new, cool features and promises improved result stability 🥳!
> > _Curious about what's new? Visit our [releases](https://github.com/CodSpeedHQ/pytest-codspeed/releases) page to delve into all the awesome details about this new version._

I gave it a try to compare on Py3.12, with `pytest-codspeed`:

- **v1.2.2**: == 1 passed, 1825 deselected in 0.24s
- **v2.2.1**: == 1 passed, 1825 deselected in 0.24s
- **v3.0.0**: == 1 passed, 1825 deselected in 4.28s:
```
┏━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━┓
┃          Benchmark ┃ Time (best) ┃ Rel. StdDev ┃ Run time ┃  Iters ┃
┡━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━┩
│ test_parse_iso8601 │   147,445ns │        4.4% │    3.00s │ 18,919 │
└────────────────────┴─────────────┴─────────────┴──────────┴────────┘
```
I don’t have this table with previous versions, not sure which value should be taken.

Secrus:
I will try to find some time to test the changes myself. I just want to be extra careful and double-check everything.

Minku-Koo:
@M5oul How can I solve problem about rust impl?
can you share me commit id?

Secrus (approved):

Secrus:
Ok, I have tested the changes locally and it seems to be alright, gonna merge it and see how it behaves with some PRs I plan on merging soon-ish.

PR #865, opened 2024-12-17
Update ruff config, use ruff format instead of black

Secrus:


Secrus:
> Do we already validate pre commit passes in ci? (I haven't looked yet, and I'm on mobile right now)

Yeah, it's part of the `Tests/Linting` pipeline

PR #871, opened 2025-01-08
Enables python 3.13

rudolfix:
This PR makes `pendulum` work on Python 3.13. We are using it a lot in our library: https://github.com/dlt-hub/dlt and initially considered forking pendulum to make it work on Python 3.13. Maybe there's a chance to merge this into main project?

I'm sure some of the changes are going too far. We are open to revert them if there's an interest from maintainers to merge this PR :)

We are running this PR with our own tests on all basic architectures and we do not see any problems

- O3 is bumped again to make it build on all Windows architectures
- More architectures are added to release build
- TimeMachine is used only in tests and present only in test deps.

happy to revert/change
- part doing the actual release commented out in github actions
- codespeed disabled (could not make it work on our bracnch)


.github/workflows/tests.yml
```diff
@@ -53,8 +53,12 @@ jobs:
...
 
       - name: Install poetry
-        run: |
-          curl -fsS https://install.python-poetry.org | python - --preview -y
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+          installer-parallel: true
+          version: 1.8.5
```
ashb, line 61: Is this change needed? I'm sure it's okay, but it does mean one extra step in the CI supply chain to worry about, so if it's not 100% strictly required, given it's not an official poetry action I'd favour sticking with just "official" curl approach?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "pendulum"
-version = "3.0.0"
+version = "3.0.1"
```
ashb, line 3: FYI: Please don't change the version in the PR like this, a sepearte release PR will make this change

ashb (commented):
Looks like something equivalent was merged in #863?

.github/workflows/tests.yml
Secrus: Since we are using github default runners, simple `pipx install poetry=={version}` is all we need.

rudolfix: No problem to revert back to curl. Should I upgrade poetry to 2.0 for the whole project? lock file is not compatible. (that's why I used the action - I could pick the right version easily)

pyproject.toml
rudolfix: sure!

rudolfix:
> Looks like something equivalent was merged in #863?

this is a continuation of #863. We had to upgrade O3 version again because we could not build Windows wheels. This again required us to handle some deprecations. I think @Secrus was able to merge the relevant commit.

We'll come back with more fixes. We will be testing updated wheels via our private PyPI and then update this PR.

.github/workflows/tests.yml
ashb: For ease of review I'd say keep the poetry upgrade/changes to a separate PR please

Secrus:
We have merged some updates to Poetry and CI setup. @rudolfix could you rebase the changes?

rudolfix:
@Secrus `pendulum` is segfaulting on alpine linux (musllinux). it is due to `mimalloc` being used (and not disabled). do you have anything against removing  mimalloc from the rust extensions? I'm not sure what are the expected speedups from using it.

also we added `any` wheel for platforms that we do not have wheels with rust built. this will disable speedups automatically. I hope that is OK. PR will come soon

Secrus:
@rudolfix I am not against it, but I don't really know the low-level semantics, so if you could TL;DR what removing it would cause, I would be grateful.

rudolfix:
@Secrus I did some investigations. in case of pendulum there's no reason to use mimalloc
https://github.com/microsoft/mimalloc?tab=readme-ov-file#benchmark-results-on-a-16-core-amd-5950x-zen3

it makes some sense on heavy multithreaded code where you allocate and deallocate on different thread. our case is exactly `cfrac` case (a lot of small allocations, single thread). so we can skip it.

Release workflow needs a lot of updates, some of actions are removed. We have a working one. I'll add this to the PR

pyproject.toml
```diff
@@ -15,12 +15,13 @@ classifiers = [
...
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
 ]
 
 dependencies = [
     "python-dateutil>=2.6",
     "tzdata>=2020.1",
-    'time-machine>=2.6.0; implementation_name != "pypy"',
+    "cffi>=1.17",
```
Secrus, line 24: What's the reason for adding `cffi`? I don't see it used in the code.

.github/workflows/release.yml
```diff
@@ -55,6 +36,35 @@ jobs:
             # mimalloc not supported on manylinux2014 cross-compile container
```
Secrus, line 36: Since we are removing the mimalloc, could you please address these comments?

Secrus (commented):
@rudolfix looks like we can drop `mimalloc`. Added some extra review comments.

ramwin:
Can we just provide a pure python version without using rust?

potiuk:
Thanks @rudolfix  (hello friend!) for raising the issue. I am eyeing now at Python 3.13 support for Airflow https://github.com/apache/airflow/pull/46891 - and of course (following @ashb involvement) - it turns out that Pendulum and Python 3.13 is not a great match for now. Putting aside the discusssion whether it works - which I have not run my tests yet on - seems that at least Pendulum **compiles** on Python 3.13 - having cargo needed to do it is for sure a pain for users to install, so having some correctness fixes but also Python 3.12 wheels is absolutely needed #879

So @Secrus @sdispater -> now there are both Airflow and dlthub people interested. Can we help somohow? Are there any blockers to make it happen? As mentioned  in #879 - we wiil shortly need to make decisions what to do with our dependencies that hold us back so it woudl be great to know if there is a way we can help to bring Pendulum up to Python 3.13 level.

Secrus:
@potiuk all the changes needed for Python 3.13 support are contained within this PR. Once it is merged, I plan on releasing Pendulum 3.1, which will include wheels for Python 3.13.

potiuk:
> @potiuk all the changes needed for Python 3.13 support are contained within this PR. Once it is merged, I plan on releasing Pendulum 3.1, which will include wheels for Python 3.13.

Fantastic news ! Thank you. If there is any thing needed - I am happy to help. I can probably very easily (providing that Github URL package installation works for pendulum) - test Pendulum from this PR in Airflow test-suite in my Python 3.13 PR. Would that be helpful?

Secrus:
@potiuk, if you have such a possibility, verifying that everything works as intended would be helpful. Thank you in advance

potiuk:
> @potiuk, if you have such a possibility, verifying that everything works as intended would be helpful. Thank you in advance

Yep - I will do some testing this weekend and come back to you :)

rust/src/python/parsing.rs
```diff
@@ -7,6 +7,7 @@ use pyo3::types::PyTime;
 use crate::parsing::Parser;
 use crate::python::types::{Duration, FixedTimezone};
 
+#[allow(deprecated)]
```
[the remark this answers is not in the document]
rudolfix: There are more deprecations in O3 0.23, I was able to fix most of them, but here the whole function must be IMO refactored. TODO added

rudolfix:
@Secrus I was able to merge our newest version and run all the CI workflows. I did all the things in your review as well. thx!

@potiuk this may be interesting to you:
We had to do a few changes to release script in order to release pendulum on all platforms. `dlt` is already using version in this PR via our private PyPI, we also released our own fork on PyPI (which we'll abandon once this is mered).
it seems that it works fine.
changes in the release script:
- build actions had to be updated, old ones were removed and stopped working. that had some consequences how files are copied into a single package
- we added `none` build that will be used on all platforms where rust build is not supported, in that build extensions are disabled
- we are using https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ for the final publish. it is easy to setup. maybe you want to change to it? if not I can paste the old step there.

potiuk:
>  we are using https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ for the final publish. it is easy to setup. maybe you want to change to it? if not I can paste the old step there.

We are in a process of switching to releases via Trusted Publishing for the whole ASF (and we are leading it, but we want to do it a lot more robust and reusable, so it takes more time) 

But I do not **really** need to publish things - I can simply build it in our CI, our CI build anyhow has all the tools to build pendulum in our CI before the .whl is published, so no worries. I even started it on weekend with  this in our hatch_build.py:


https://github.com/apache/airflow/pull/46891/files#diff-d56d1b58c0ca28bdddd10c2cee2ab3cb4312766ea5de51f210dbc15a0fea42faR403

```
"pendulum @ git+https://github.com/python-pendulum/pendulum.git@76422468575b5817f7a58f3d05c26b480bf671a9"
```

But I had other issues with the build and other things to look at - so I am only on it now :)

potiuk:
Yeah - https://github.com/apache/airflow/pull/46891 and while it will still have some failures, I hope we will see if there will be any related to pendulum - but yeah - once I have cargo in my image (I have because some other packages for Python 3.13 need it) - Pendulum nicely installs with github-url installatoin.

rudolfix:
@Secrus is there anything that could help you merge this?

Secrus:
> @Secrus is there anything that could help you merge this?

I would appreciate it if you could make the days longer, so I can do more in one day :rofl: 

On a more serious note, I will dive into this PR (hopefully) this week and see where we are.

rudolfix:
> > @Secrus is there anything that could help you merge this?
> 
> I would appreciate it if you could make the days longer, so I can do more in one day 🤣
> 
you should move to south pole then :)

rust/src/python/helpers.rs
```diff
@@ -76,7 +76,7 @@ impl PartialOrd for DateTimeInfo<'_> {
 pub fn get_tz_name<'py>(dt: &Bound<'py, PyAny>) -> PyResult<String> {
     // let tz: &str = "";
 
-    if !PyDateTime::is_type_of_bound(dt) {
+    if !PyDateTime::is_type_of(dt) {
```
Secrus, line 79: Since `dt` is of type `Bound`, shouldn't that be as it was? (I am not yet versed in the pyo3 stuff, so I might be wrong.)

Secrus (commented):
@rudolfix Just one question, otherwise looks good to me.

rust/src/python/helpers.rs
Secrus: Nevermind, I read some more on the matter, and it seems PyO3 is consolidating and removing the `*_bound` methods.

Secrus (approved):
@rudolfix, it seems like the only thing needed for this PR to be merged is a rebase. Huge thanks for all your hard work on this matter.

PR #890, opened 2025-04-16
Fix sed in release pipeline.

ashb:
The `;` was missing between the two `s///` statements; without it it failed
with an error of `sed: -e expression #1, char 86: unknown option to `s'`

Tested with act to get _some_ sembalance of an idea if it will work, but it's
no guarantee

```
act \
  --env GITHUB_REPOSITORY=python-pendulum/pendulum \
  -P ubuntu-latest=ghcr.io/catthehacker/ubuntu:rust-latest \
  --artifact-server-path $PWD/.artifacts \
  -W .github/workflows/release.yml \
  --matrix container:off,target:aarch64 \
  --container-architecture linux/amd64
```

## Pull Request Check List

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -103,7 +103,7 @@ jobs:
       - name: Hotswap build backend for Poetry
         # Maturin doesn't support building no-extension wheels, so we swap to Poetry for that
         run: |
-          sed -i '/^\[build-system\]/,/^\[/{s/^requires = .*/requires = ["poetry-core>=2.0.0,<3.0.0"]/ s/^build-backend = .*/build-backend = "poetry.core.masonry.api"/}' pyproject.toml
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       - name: Hotswap build backend for Poetry
         # Maturin doesn't support building no-extension wheels, so we swap to Poetry for that
         run: |
-          sed -i '/^\[build-system\]/,/^\[/{s/^requires = .*/requires = ["poetry-core>=2.0.0,<3.0.0"]/ s/^build-backend = .*/build-backend = "poetry.core.masonry.api"/}' pyproject.toml
+          sed -i -e '/^\[build-system\]/,/^\[/{s/^requires = .*/requires = ["poetry-core>=2.0.0,<3.0.0"]/; s/^build-backend = .*/build-backend = "poetry.core.masonry.api"/}' pyproject.toml
```
ashb, line 106: The `-e` may or may have not been required, but this def worked for me in `act`

Secrus (approved):

PR #897, opened 2025-04-22
Deprecate `__version__` attribute

Secrus:
## Pull Request Check List

### DIFF
--- a/src/pendulum/__init__.py
+++ b/src/pendulum/__init__.py
@@ -2,11 +2,11 @@ from __future__ import annotations
 
 import datetime as _datetime
 
+from typing import Any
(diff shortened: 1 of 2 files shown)


ashb (approved):
I kind of like having `__version__` attrs in there, as `importlib.metadata` is surprisingly slow, but I guess not much should really be looking at version anyway

Secrus:
> I kind of like having `__version__` attrs in there, as `importlib.metadata` is surprisingly slow, but I guess not much should really be looking at version anyway

Looking at version shouldn't be something one does too often, and I prefer we have a single (or rather reduce from 3 to 2) source of version.

PR #902, opened 2025-06-05
feat: add Hindi (hi) localization support

gjaynir0508:
# Add Localization Support for Hindi (`hi`) (Indian Language)
This PR adds Hindi (hi) locale support for Pendulum's date and time formatting.

Includes:
- Full translations for relative time phrases (`ago`, `in`, `before`, `after`)
- Abbreviated and full month/day names
- Date/time format definitions (L, LL, etc.)
- Patterns similar to existing locales

Note:
Hindi typically does not use ordinal indicators (like "1st", "2nd") in standard date or time expressions. 
As such, this PR omits ordinal mappings (`Do`, `Qo`, etc.) for now.

If needed in future use cases (like grammatical sentence generation or quarter formatting), a limited ordinal map can be added later.

Tested against typical `diff_for_humans()` and formatting scenarios.

Let me know if any adjustments are needed!

## Pull Request Check List

- [ ] ~Updated **documentation** for changed code.~ (Not Applicable)

### DIFF
--- a/tests/localization/test_hi.py
+++ b/tests/localization/test_hi.py
@@ -13,58 +13,57 @@ def test_diff_for_humans():
 
 def diff_for_humans():
     d = pendulum.now().subtract(seconds=1)
-    assert d.diff_for_humans(locale=locale) == 'कुछ सेकंड पहले'
+    assert d.diff_for_humans(locale=locale) == "कुछ सेकंड पहले"
 
     d = pendulum.now().subtract(seconds=2)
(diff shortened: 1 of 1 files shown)


gjaynir0508:
Note: I noticed that the `clock` utility currently doesn't specify encoding when opening files. 
For certain environments (especially on Windows) this can cause issues when reading locale files that contain Unicode (e.g., Hindi).

While it's not included in this PR to keep the scope limited to locale data, I suggest updating the `open()` calls in the clock utility to:

```python
open(path, "w", encoding="utf-8")
```

PR #922, opened 2025-10-22
Bump PyO3 version to 0.27

botantony:
## Pull Request Check List

This will add Python 3.14 support. I was able to run `make dev && make test` successfully (no errors, only 3 warnings and 4 skips) so I assume there was no breaking changes for this project

Related to https://github.com/Homebrew/homebrew-core/pull/250674

### DIFF
--- a/rust/src/python/helpers.rs
+++ b/rust/src/python/helpers.rs
@@ -100,7 +100,7 @@ pub fn get_tz_name<'py>(dt: &Bound<'py, PyAny>) -> PyResult<String> {
     }
 
     if let Some(tzname_attr) = tzname_attr {
-        let tzname: &Bound<PyString> = tzname_attr.downcast()?;
+        let tzname: &Bound<PyString> = tzname_attr.cast()?;
         tzname.extract()
     } else {
         Ok(String::new())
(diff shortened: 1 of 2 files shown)


ashb:
Looks like we've got some deprecated things to fix now too:

> error: use of deprecated method `pyo3::types::PyAnyMethods::downcast`: use `Bound::cast` instead

ashb:
We should also add py 3.14 to the test matrix (and likely remove 3.9 since that is out of support)

pyproject.toml
```diff
@@ -15,6 +15,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
```
Secrus: Drop that please, let's make this PR focused on what the title says, upgrading PyO3 0.27.

Secrus (changes requested):
Please address the deprecations and changes between 0.25 and 0.27.

pyproject.toml
ashb: Yeah sorry, that's a fair comment.

botantony: Sorry, missed the review, it should be removed now

ashb (approved):

PR #930, opened 2026-01-09
Drop python 3.9 support

mattwang44:
Drop python 3.9 support as it has reached EOL


mattwang44:
> Please drop the changes to CHANGELOG. Also rebase the changes to CI files

@Secrus Thanks for the review. PTAL.

PR #957, opened 2026-04-03
Make Ukrainian locale name follow ISO 639, update it and add tests

### LINKED ISSUE #955: Ukrainian locale should be uk not ua

- **OS version and name**: Alpine Linux 3.24.0_alpha20260127
- **Pendulum version**: 3.2.0

## Issue
Hello! PR #793 added Ukrainian localization with the code `ua`, but the ISO 639 language code, not to be confused with the ISO 3166 country code, is `uk` for the Ukrainian language, e.g. in GNU/Linux the locale is `uk_UA.UTF-8`, and browsers send headers like `Accept-Language: uk-UA,uk` to servers. What would be the correct way to make Pendulum use the correct `uk` language code, with a fallback to `ua` for backward compatibility?

```
>>> import locale
>>> import pendulum
>>> locale.getlocale()
('uk_UA', 'UTF-8')
>>> locale.getlocale()[0].split('_')[0]
'uk'
>>> pendulum.now('Europe/Kyiv').diff_for_humans(locale='uk')
  File "/home/user/pendulum/src/pendulum/datetime.py", line 746, in diff_for_humans
    return pendulum.format_diff(diff, is_now, absolute, locale)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/pendulum/src/pendulum/helpers.py", line 171, in format_diff
    return _difference_formatter().format(diff, is_now, absolute, locale)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/pendulum/src/pendulum/formatting/difference_formatter.py", line 52, in format
    locale = self._locale if locale is None else Locale.load(locale)
                                                 ^^^^^^^^^^^^^^^^^^^
  File "/home/user/pendulum/src/pendulum/locales/locale.py", line 40, in load
    raise ValueError(f"Locale [{locale}] does not exist.")
ValueError: Locale [uk] does not exist.
>>> pendulum.now('Europe/Kyiv').diff_for_humans(locale='ua')
'кілька секунд тому'
```

nykula:
## Pull Request Check List

Fixes #955.


tests/localization/test_uk.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_diff_for_humans():
+    with pendulum.travel_to(pendulum.datetime(2026, 4, 3), freeze=True):
+        diff_for_humans()
+
+
+def diff_for_humans():
+    d = pendulum.now().subtract(seconds=1)
+    assert d.diff_for_humans(locale=locale) == "кілька секунд тому"
```
afastronomer, line 16: assert d.diff_for_humans(locale=locale) == "1 секунду тому"

src/pendulum/locales/uk/locale.py
```diff
@@ -0,0 +1,257 @@
...
+
+"""
+uk locale file.
+
+It has been generated automatically and must not be modified directly.
+"""
+
+
+locale = {
+    'plural': lambda n: 'few' if (((0 == 0 and ((0 == 0))) and ((n % 10) == (n % 10) and (((n % 10) >= 2 and (n % 10) <= 4)))) and (not ((n % 100) == (n % 100) and (((n % 100) >= 12 and (n % 100) <= 14))))) else 'many' if ((((0 == 0 and ((0 == 0))) and ((n % 10) == (n % 10) and (((n % 10) == 0)))) or ((0 == 0 and ((0 == 0))) and ((n % 10) == (n % 10) and (((n % 10) >= 5 and (n % 10) <= 9))))) or ((0 == 0 and ((0 == 0))) and ((n % 100) == (n % 100) and (((n % 100) >= 11 and (n % 100) <= 14))))) else 'one' if (((0 == 0 and ((0 == 0))) and ((n % 10) == (n % 10) and (((n % 10) == 1)))) and (not ((n % 100) == (n % 100) and (((n % 100) == 11))))) else 'other',
```
ashb, line 12: There are an awful lot of `0 == 0` in here. More than zero is too many

ashb (dismissed):

tests/localization/test_uk.py
```diff
@@ -0,0 +1,119 @@
...
+
+    d = pendulum.now().subtract(days=2)
+    assert d.diff_for_humans(locale=locale) == "2 дні тому"
+
+    d = pendulum.now().subtract(weeks=1)
+    assert d.diff_for_humans(locale=locale) == "1 тиждень тому"
+
+    d = pendulum.now().subtract(weeks=2)
+    assert d.diff_for_humans(locale=locale) == "2 тижні тому"
+
```
afastronomer, line 65: d = pendulum.now().subtract(weeks=5)
    assert d.diff_for_humans(locale=locale) == "5 тижнів тому"

tests/localization/test_uk.py
```diff
@@ -0,0 +1,119 @@
...
+    d = pendulum.now().subtract(hours=5)
+    assert d.diff_for_humans(locale=locale) == "5 годин тому"
+
+    d = pendulum.now().subtract(days=1)
+    assert d.diff_for_humans(locale=locale) == "1 день тому"
+
+    d = pendulum.now().subtract(days=2)
+    assert d.diff_for_humans(locale=locale) == "2 дні тому"
+
```
afastronomer, line 56: d = pendulum.now().subtract(days=5)
    assert d.diff_for_humans(locale=locale) == "5 днів тому"

tests/localization/test_uk.py
nykula: Pendulum considers diffs less than 10 seconds, including 0 or 1, to be "a few seconds", so I'd leave this test as is, which matches the behavior of other locales.

src/pendulum/locales/uk/locale.py
nykula: Agreed that checking zero against zero is useless. But that part comes from `clock`, which generates the condition from Babel data? Such generated checks are also present in `cs`, `pl`, `sk` locales. I'm open to trying to fix this, but can you suggest how I should approach it?

tests/localization/test_uk.py
nykula: `weeks=5` behaves in the same way as `months=1`, resulting in `1 місяць тому`. Should I add that assertion?

tests/localization/test_uk.py
nykula: Makes sense to check, thanks, will commit when I have more time.

src/pendulum/locales/uk/locale.py
ashb: Oh it's generated (and in the previous version too). Leave it in this pr

tests/localization/test_uk.py
ashb: What do other locales do here?

nykula: All locales behave the same in this regard. @afastronomer most likely meant that from the grammar point of view, 2 weeks and 5 weeks correspond to different forms in Ukrainian. But the latter form is currently unused in Pendulum; there's no feature, that I know of, to get a human representation of a duration in a specific unit such as weeks. One way to get to the form is:

```py
pendulum.locale('uk').translation(f"units.week.{pendulum.locale('uk').plural(5)}").format(5)
# -> '5 тижнів'
```

src/pendulum/locales/uk/locale.py
Secrus: Let's leave this as is. We can look into refactoring the `clock` script and fixing the weird logic at a later date.

src/pendulum/locales/locale.py
```diff
@@ -53,7 +53,11 @@ def normalize_locale(cls, locale: str) -> str:
         if m:
             return f"{m.group(1).lower()}_{m.group(2).lower()}"
         else:
-            return locale.lower()
+            lower = locale.lower()
+            if lower == 'ua':
+                # Backward compatibility. Ukrainian was 'ua' initially.
+                lower = 'uk'
+            return lower
```
Secrus, line 60: Don't think we need that, since the initial locale had the wrong code. @ashb what do you think?

nykula: So, do I remove this backward-compatibility part or keep it? I'm fine with either decision.

Secrus: Let's remove it

nykula: Done 👌

nykula:
My bad, left the updated test file uncommitted.

Secrus (approved):

PR #960, opened 2026-04-10
fix: accept naive stdlib datetime in DateTime.diff()

### LINKED ISSUE #880: naive DateTime.diff fails when the input is a naive datetime object, as pendulum.instance converts it to a pendulum.DateTime with UTC tz

- **OS version and name**: Ubuntu 22.04.3 LTS (running in WSL2)
- **Pendulum version**:  3.0.0

## Issue

When a naive pendulum.DateTime calls `diff` with a stdlib datetime object the input is converted to a pendulum instance, but it defaults to having a timezone of UTC, causing a TypeError

`TypeError: can't compare offset-naive and offset-aware datetimes`

Repro

```python
from datetime import datetime

import pendulum

pendulum_today = pendulum.today().naive()
stdlib_today = datetime.today()

print(pendulum_today.tzinfo)
print(stdlib_today.tzinfo)

pendulum_today.diff(stdlib_today)
```

JKDasondee:
Fixes #880

`DateTime.diff()` raises `TypeError` when called with a naive stdlib `datetime` object on a naive `DateTime`. The fix wraps the stdlib datetime with `pendulum.instance(dt, tz=None)` to preserve naivety before passing to `Interval`.

### DIFF
--- a/src/pendulum/datetime.py
+++ b/src/pendulum/datetime.py
@@ -709,6 +709,13 @@ class DateTime(datetime.datetime, Date):
         if dt is None:
             dt = self.now(self.tz)
 
+        if (
+            isinstance(dt, datetime.datetime)
+            and not isinstance(dt, DateTime)
+            and dt.tzinfo is None
+        ):
+            dt = pendulum.instance(dt, tz=None)
+
         return Interval(self, dt, absolute=abs)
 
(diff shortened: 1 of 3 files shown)


Secrus:
This is some weird mixture of commits. Please, clean up your work and provide tests.

PR #984, opened 2026-07-09
Fix lowercase 'a' token formatting to lowercase meridiem

vidigoat:
The `a` format token returns the literal string `a` instead of a lowercase meridiem indicator (am/pm).

```python
>>> import pendulum
>>> dt = pendulum.datetime(2016, 8, 28, 23)
>>> dt.format("A")
'PM'
>>> dt.format("a")   # expected 'pm'
'a'
```

`_format_localizable_token` only handles the uppercase `A` token, so `a` falls through to the default branch that returns the token unchanged. The `a` token is documented and registered in the token regex, and the parser side already treats it as a lowercase meridiem, so formatting and parsing were inconsistent and a value formatted with `a` could not be round-tripped:

```python
>>> pendulum.from_format(dt.format("YYYY-MM-DD hh:mm a"), "YYYY-MM-DD hh:mm a")
```

previously failed because the formatted string contained a literal `a` rather than `pm`.

This mirrors the existing `A` branch and lowercases the result for the `a` token. Added a regression test alongside the existing `test_am_pm`.

### DIFF
--- a/docs/docs/string_formatting.md
+++ b/docs/docs/string_formatting.md
@@ -140,6 +140,7 @@ The following tokens are currently supported:
 |                                | SSSS ...      | 000[0..] 001[0..] ... 998[0..] 999[0..]    |
 |                                | SSSSSS        |                                            |
 | **AM / PM**                    | A             | AM, PM                                     |
+|                                | a             | am, pm                                     |
 | **Timezone**                   | Z             | -07:00, -06:00 ... +06:00, +07:00          |
 |                                | ZZ            | -0700, -0600 ... +0600, +0700              |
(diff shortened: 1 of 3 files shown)


vedaant00:
I hit this same bug independently and opened #990 before noticing yours, so I'm closing mine as a duplicate since this one came first.

One addition that might help it land: the format-token docs table in `docs/docs/string_formatting.md` lists `A` but not `a`, so it's worth adding an `a` row there too:

```
|                                | a             | am, pm                                     |
```

Happy to push that to your branch if useful, or you can grab it from #990.

PR #990, opened 2026-07-28
Fix `a` format token returning literal instead of lowercase meridiem

vedaant00:
### What

`DateTime.format("a")` returns the literal string `"a"` instead of the lowercase am/pm marker:

```python
import pendulum
dt = pendulum.datetime(2020, 1, 1, 14, 0, 0)
dt.format("A")  # "PM"  (correct)
dt.format("a")  # "a"   (bug — expected "pm")
```

### Why

The `a` token is declared in the formatter regex and in `_LOCALIZABLE_TOKENS`, but `_format_localizable_token` has a branch for `A` and none for `a`, so `a` falls through to `return token`.

### Fix

- Add the missing `a` branch, mirroring `A` and lowercasing the localized day period (matching the existing `_LOCALIZABLE_TOKENS["a"]` lambda). Non-Latin locales like `ja` are unaffected since `.lower()` is a no-op there.
- Added `test_am_pm_lowercase`.
- Added the `a` row to the format-token docs table.

Full suite passes (1838 passed).

## Pull Request Check List


vedaant00:
Closing as a duplicate of #984, which predates this by a few weeks. Left a note there about also adding the `a` row to the format-token docs table (the one extra thing this PR covered).

PR #1001, opened 2026-09-01
Fix KeyError in diff_for_humans for the zh locale

dylanpulver:
## Pull Request Check List

---

`diff_for_humans()` raises `KeyError: 'time'` for the `zh` locale whenever it is given a
second datetime (the non-relative-to-now form), for every unit:

```python
>>> a = pendulum.datetime(2024, 6, 15, 12, 0, 5)
>>> b = pendulum.datetime(2024, 6, 15, 12, 0, 0)
>>> a.diff_for_humans(b, locale="zh")
KeyError: 'time'
>>> a.diff_for_humans(b, locale="ja")
'数秒 後'
```

`zh/custom.py:9-10` holds `"after": "{time}后"` / `"before": "{time}前"` — named placeholders —
while `DifferenceFormatter.format()` substitutes positionally
(`difference_formatter.py:115` and `:159`, both `locale.get(key).format(time)`). All 28 other
locales use `{0}`, so this changes only the `zh` data; no code change.

Base is `master` per the template note (bug fix, not a behaviour change).

**Why nothing caught it:** the relative-to-now form reads `translations.relative.*` (the
generated CLDR table) and never touches these two custom entries, and there is no
`tests/localization/test_zh.py` — `zh`, `en*` and `pt_br` are the locales with no test file.
Added one for `zh`, modelled on `test_ja.py`, covering both forms.

Found by a sweep, not a bug report: `diff_for_humans` across all 29 locales x 8 durations x
both signs x `absolute` on/off (928 calls), looking for exceptions or unsubstituted `{`. `zh`
was the only locale that failed, 16/32 of its calls — exactly the `is_now=False,
absolute=False` path. `{time}` appears nowhere else in `src/` or `tests/`.

**Measured:** `pytest tests` 1840 passed / 3 skipped before, 1841 / 3 after (same command,
same venv). Mutants: reverting the two strings and keeping the new test re-raises all 16
sweep failures and fails `test_zh.py`.

**Honest limitation:** a second mutant — keeping `{time}` and making the formatter call
`.format(time, time=time)` — also passes the sweep and the new test, so the test does not by
itself pick between the two remedies. I chose the data fix because the alternative sanctions
two placeholder conventions in the locale files; say so and I'll switch it.

**Not tested:** I did not add test files for `en`, `en_gb`, `en_us` or `pt_br`, which also
have none. `ruff format --check` reports a pre-existing blank-line diff in 28 of the 29
`locales/*/custom.py` files including this one; I left it alone.

*Disclosure: the diff, the test and this description were drafted with Claude Opus 5.*


tests/localization/test_zh.py
```diff
@@ -0,0 +1,45 @@
...
+
+locale = "zh"
+
+
+def test_diff_for_humans():
+    with pendulum.travel_to(pendulum.datetime(2016, 8, 29), freeze=True):
+        diff_for_humans()
+
+
+def diff_for_humans():
```
Secrus: why is this a separate function instead of putting it inside context manager in the test function?

Secrus (approved):

PR #1009, opened 2026-09-12
Restore hashing for named timezones

### LINKED ISSUE #1008: Timezones are not hashable

- **OS version and name**: Ubuntu 24.04.4 LTS
- **Pendulum version**: 3.2.0

## Issue

I noticed **Pendulum `Timezone` objects are not hashable** since version 3.1.0 but they were in version 3.0.0.

## Minimal reproducible example (versions 3.1.0 and 3.2.0)
```
Python 3.11.8 (main, Jul 25 2025, 15:41:07) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pendulum
>>> hash(pendulum.Timezone("UTC"))
TypeError: unhashable type: 'Timezone'
```

## Expected behavior (version 3.0.0)
```
Python 3.11.8 (main, Jul 25 2025, 15:41:07) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pendulum
>>> hash(pendulum.Timezone("UTC"))
8423064200767
```

I believe this is a side-effect of this commit https://github.com/python-pendulum/pendulum/commit/6383c07 where `__eq__` was defined, but **`__hash__` wasn't**, breaking Python's definition of [hashable](https://docs.python.org/3.13/glossary.html#term-hashable).

In version 3.0.0 (and _probably_ in earlier versions as well) the `Timezone` class inherited both methods, resulting in the class being hashable.

Since I think non-hashability wasn't intended, I'm reporting this as a bug. 

Hopefully I'll be able to implement a fix myself, but if not, I hope someone else can.

Thank you!

emme1t:
Fixes #1008.

Adding `Timezone.__eq__` in 3.1.0 made named timezone objects unhashable. This breaks calls such as `hash(pendulum.Timezone("UTC"))` and using a timezone as a dictionary key or set element.

Implement `__hash__` using the timezone key, matching the existing equality rule. The regression tests construct equal objects with `Timezone.no_cache()` and verify equal hashes, dictionary lookup, and set deduplication. A separate test checks dictionary lookup for two distinct timezone names.

## Validation

Tested locally on Windows with Python 3.13.13:

- All four new regression cases fail on the original implementation.
- Timezone suite: **638 passed, 5 skipped**.
- Complete pure-Python suite (`PENDULUM_EXTENSIONS=0`): **1,840 passed, 5 skipped, 2 failed**. A clean checkout of the original commit produces the same two failures, with 1,836 passing tests. The failing tests are `tests/date/test_diff.py::test_diff_for_humans_now_and_future_month` and `tests/test_parsing.py::test_parse`.
- The project's pinned Ruff 0.14.11 lint and format checks pass for both changed files.
- The lockfile's Mypy 1.13.0 passes for all 223 source/test files.
- `git diff --check` passes.

The native extension build could not run on this machine because the MSVC `link.exe` linker is unavailable. The Python source tests use the same pure-Python mode exercised by the repository's CI.

## Pull Request Check List

- [x] Added tests for changed code.
- [ ] Updated documentation for changed code (existing behavior is restored; no user-facing option changes).

This patch, regression tests, and PR description were prepared with OpenAI Codex assistance. The reported reproductions and tests were executed locally.


src/pendulum/tz/timezone.py
```diff
@@ -69,6 +69,9 @@ def __new__(cls, key: str) -> Self:
     def __eq__(self, other: object) -> bool:
         return isinstance(other, Timezone) and self.key == other.key
 
+    def __hash__(self) -> int:
+        return hash(self.key)
```
ashb: This is slightly incorrect, as it means that `"UTC"` and `Timezone("UTC")` would hash to the same value

emme1t: Thanks for pointing this out. Updated in b623446 to hash `(Timezone, self.key)`, giving named timezones a type-qualified hash input. Using the base `Timezone` class also preserves equal hashes for subclasses that inherit the existing equality behavior.

Added regression coverage for equal base/subclass instances as dictionary keys and set members across UTC, Europe/Paris, and America/New_York.

Local validation on Windows / CPython 3.13 in pure-Python mode: 641 timezone tests passed, 5 skipped. The full suite has 1,843 passed, 5 skipped, and the same two failures reproduced on the original upstream checkout (`test_diff_for_humans_now_and_future_month` and `test_parse`). Mypy 1.13.0 passes across 223 files. The Rust extension was not validated.

Ruff 0.14.11 lint and format checks pass as well. This patch and validation were prepared with Codex assistance.

erovira: > "UTC" and Timezone("UTC") would hash to the same value

Hey @ashb . Curious, in what sense is the initial (simpler) implementation incorrect? From that I understand, the rule is that if objects compare equal, then their hashes need to be equal as well. 

But AFAIK there's no problem with hashes colliding if the objects are not equal, which is the case of the `"UTC"` string object and the `Timezone("UTC")` object.

